PR #11, opened 2025-03-30
fix: linting

jmeridth:
- [x] use latest golangci-lint version (v2.0.2)
- [x] fix new linting issues (errors in defer)


eddie-knight (approved):

eddie-knight:
```sh
Error: invalid version string 'v2.0.2', golangci-lint v2 is not supported by golangci-lint-action v6, you must update to golangci-lint-action v7.
```

PR #15, opened 2025-04-03
test: adding unit testing for layer4.Result#UpdateAggregateResult func

trumant:


eddie-knight (approved):
Ahhhhhh, thank you for this!

PR #19, opened 2025-04-15
feat: layer 4 schema changes for discussion

trumant:
Understanding the intent and reasoning behind schema choices is
difficult without clear documentation. All of the definitions and
fields in the layer4 schema are now documented and the example data
file in `schemas/example_evaluation_results.yml` should help
further explain the schema.

 * snake case is adopted for all field names
 * Struct disjunction is introduced to correctly specify a constraint between the execution status of AssessmentMethod and the presence of a result
 * List bounds are refined to more correctly specify the number of expected items and the expected type constraints for those items
 * `@Go` annotations are added to show the desired/resulting Go type/field names where needed to cope with snake cased field names

Once these changes are merged, we should be unblocked to generate Go types
from the cue schema and hopefully, also generate documentation from the
comments within the schema.

### DIFF
--- a/schemas/example_evaluation_results.yml
+++ b/schemas/example_evaluation_results.yml
@@ -1,8 +1,9 @@
-frameworkID: OSPS
+catalog_id: OSPS
+corrupted_state: false
 evaluations:
-  - controlID: "OSPS-BR-01"
+  - control_id: "OSPS-BR-01"
     assessments:
-      - requirementID: "OSPS-BR-01.01"
+      - requirement_id: "OSPS-BR-01.01"
         methods:
           - name: "Manual review of CI pipelines used by the project"
             description: "Manually review the CI pipelines for proper handling of user-controlled build inputs."
@@ -15,7 +16,7 @@ evaluations:
             run: true
             result:
               status: needs_review
-      - requirementID: "OSPS-BR-01.02"
+      - requirement_id: "OSPS-BR-01.02"
         methods:
           - name: "Manual review of CI pipelines used by the project"
             description: "Manually review the CI pipelines for proper handling of branch-name build inputs."
--- a/schemas/example_evaluation_results.yml
+++ b/schemas/example_evaluation_results.yml
@@ -1,8 +1,6 @@
-name: Open Source Security Baseline
 frameworkID: OSPS
 evaluations:
-  - name: "OSPS BR Controls Evaluation"
-    controlID: "OSPS-BR-01"
+  - controlID: "OSPS-BR-01"
     assessments:
       - requirementID: "OSPS-BR-01.01"
         methods:
--- a/schemas/example_evaluation_results.yml
+++ b/schemas/example_evaluation_results.yml
@@ -1,0 +1,34 @@
+name: Open Source Security Baseline
+frameworkID: OSPS
+evaluations:
+  - name: "OSPS BR Controls Evaluation"
+    controlID: "OSPS-BR-01"
+    assessments:
+      - requirementID: "OSPS-BR-01.01"
+        methods:
+          - name: "Manual review of CI pipelines used by the project"
+            description: "Manually review the CI pipelines for proper handling of user-controlled build inputs."
+            run: false
+          - name: "One time SAST scan of GitHub Actions workflows"
+            description: "GitHub Actions-specific SAST scan run against all contents of `.github/workflows`."
+            run: false
+          - name: "Ongoing CodeQL SAST scanning of GitHub Actions workflows"
+            description: "Project is configured to run regular CodeQL SAST scans of GitHub Actions workflows."
+            run: true
+            result:
+              status: needs_review
+      - requirementID: "OSPS-BR-01.02"
+        methods:
+          - name: "Manual review of CI pipelines used by the project"
+            description: "Manually review the CI pipelines for proper handling of branch-name build inputs."
+            run: false
+          - name: "One time SAST scan of GitHub Actions workflows"
+            description: "GitHub Actions-specific SAST scan run against all contents of `.github/workflows`."
+            run: true
+            result:
+              status: passed
+          - name: "Ongoing CodeQL SAST scanning of GitHub Actions workflows"
+            description: "Project is configured to run regular CodeQL SAST scans of GitHub Actions workflows."
+            run: true
+            result:
+              status: passed
--- a/schemas/layer-4.cue
+++ b/schemas/layer-4.cue
@@ -21,9 +21,6 @@ package schemas
 
 // ControlEvaluation describes the evaluation of the layer 2 control referenced by controlID and the assessment of that control's requirements.
 #ControlEvaluation: {
-    // name of the control being evaluated
-    // TODO: why materialize/duplicate this data here if a consumer can simply reference the layer 2 data?
-    name: string
     // ID of the layer 2 control being evaluated
     controlID: string
     // TODO: should there also be a frameworkID here to make a ControlEvaluation more self-contained?
@@ -50,7 +47,7 @@ package schemas
     description: string
     // Run is a boolean indicating whether the method was run or not. When run is true, result is expected to be present.
     run: bool
-    // Remediation guide is a ¯\_(ツ)_/¯
+    // Remediation guide is a URL to remediation guidance associated with the control's assessment requirement and this specific assessment method.
     remediation_guide?: #URL
 }
 
@@ -69,11 +66,12 @@ package schemas
 }
 
 // AssessmentResult describes the result of the assessment of a layer 2 control requirement.
-//  * passed when all evidence suggests the control is met
-//  * failed when some evidence suggests the control is not met
-//  * needs_review when evidence was gathered but cannot be reliably interpreted to reach a decision. A human should review the evidence gathered
-//  * error when the method failed to execute
 #AssessmentResult: {
+    // status describes what happened when the assessment method was run
+    //  * passed when all evidence suggests the control is met
+    //  * failed when some evidence suggests the control is not met
+    //  * needs_review when evidence was gathered but cannot be reliably interpreted to reach a decision. A human should review the evidence gathered
+    //  * error when the method failed to execute
     status: #Status
     // TODO: I can imagine assessment methods potentially making more than a single change, perhaps this should be a list
     change?: #Change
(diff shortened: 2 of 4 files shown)


schemas/layer-4.cue
```diff
@@ -1,35 +1,45 @@
...
 
+#URL: =~"^https?://[^\\s]+$"
+
 #ControlEvaluation: {
+    // name of the control being evaluated
     name: string
-    "control-id": string
-    result: #Result
-    message: string
-    "documentation-url"?: =~"^https?://[^\\s]+$"
```
trumant, line 14: maybe intended for "future planned use"

schemas/layer-4.cue
```diff
@@ -1,35 +1,45 @@
...
-    "corrupted-state"?: bool
-    "assessment-results"?: [...#AssessmentResult]
+    // ID of the control being evaluated
+    controlID: string
+    // one or more assessments of the control requirements
+    assessments: [#Assessment, ...]
```
trumant: This seems extremely non-controversial, just fixing the schema to specify that we expect 1 or more assessments

schemas/layer-4.cue
trumant: What would the content at the value of `documentation-url` describe/document? As it is optional, until we have more clarity there, I propose we remove it.

schemas/layer-4.cue
```diff
@@ -1,35 +1,45 @@
...
 // Types
 
+#URL: =~"^https?://[^\\s]+$"
+
 #ControlEvaluation: {
+    // name of the control being evaluated
     name: string
-    "control-id": string
-    result: #Result
-    message: string
```
trumant, line 13: In the absence of documentation or existing usage, `message` should go away.

Reviewers - Please provide any such examples if you are aware of them

schemas/layer-4.cue
```diff
@@ -1,35 +1,45 @@
...
+    methods: [#AssessmentMethod, ...]
+}
+
+#AssessmentMethod: {
+    // the method used to assess the requirement
     name: string
+    // detailed explanation of the method
     description: string
-    message: string
-    "function-address" string
```
trumant: In the absence of documentation or existing usage, `function-address` should go away.

schemas/layer-4.cue
```diff
@@ -1,35 +1,45 @@
...
-    change?: #Change
-    value?: _
```
trumant, line 26: In the absence of documentation or existing usage, `value` should go away.

schemas/layer-4.cue
```diff
@@ -1,35 +1,45 @@
...
+    //  * failed when some evidence suggests the control is not met
+    //  * needs_review when evidence was gathered but cannot be reliably interpreted to reach a decision. A human should review the evidence gathered
+    //  * error when the method/strategy failed to execute
+    result?: "passed" | "failed" | "needs_review | error"
+    remediation_guide?: #URL
 }
-
-#Result: "Passed" | "Failed" | "Needs Review"
-
-#Change: {
```
trumant, line 31: There is suspected existing usage of this definition/type in a currently private repository that I do not have access to. Until I can see that usage and understand it better, this specific removal should be considered for discussion only.

schemas/example_evaluation_results.yml
```diff
@@ -0,0 +1,34 @@
+name: Open Source Security Baseline
```
trumant: This example file was validated by running: `cue vet schemas/layer-4.cue schemas/example_evaluation_results.yml -d '#Evaluation'`

schemas/layer-4.cue
```diff
@@ -1,35 +1,104 @@
 // Top level schema //
 
-"evaluation-plans": [...#EvaluationPlan]
+// Framework Evaluation describes the evaluation of a single framework and its controls.
+#FrameworkEvaluation: {
+    // name of the framework being evaluated, ex: NIST 800-53, CCC, OSPS, etc.
+    // TODO: why materialize/duplicate this data here if a consumer can simply reference the layer 2 data?
```
trumant: More specifically, should we remove the `name` field altogether from this definition as it is a property of the framework name, not of a FrameworkEvaluation.

schemas/layer-4.cue
```diff
@@ -1,35 +1,104 @@
...
+    name: string
+    // frameworkID contains the unique identifier of the framework being evaluated
+    // TODO: in the case of versioned frameworks (ex: NIST 800-53), should we expect the version to be part of the ID?
+    frameworkID: string
```
trumant: Frameworks are a layer 1 concern for which no schema exists, but I think it's safe to imagine that they will have unique identifiers.

schemas/layer-4.cue
```diff
@@ -1,35 +1,104 @@
...
 
-// Types
+// URL describes a specific subset of URLs of interest to the framework
+// TODO: this definition should be imported from a more appropriate module/package
+#URL: =~"^https?://[^\\s]+$"
 
+// ControlEvaluation describes the evaluation of the level 2 control referenced by controlID and the assessment of that control's requirements.
 #ControlEvaluation: {
+    // name of the control being evaluated
+    // TODO: why materialize/duplicate this data here if a consumer can simply reference the layer 2 data?
```
trumant: Specifically, the `name` field here is a property of the layer 2 control, not of the `ControlEvaluation` and therefore I'd like to remove it from this definition and rely on consumers following the reference to the layer2 control data if they need the name.

trumant:
https://github.com/trumant/sci/pull/1 shows how these changes will be used to address go generation

schemas/layer-4.cue
jpower432: I agree with this. I think a unique id for the framework would make it easier for tooling to reference the source control set and create targeted plans.

trumant:
@jpower432 Appreciate your review!

jpower432:
> @jpower432 Appreciate your review!

Happy to provide input @trumant! Your code comments were really helpful for me. A general question, I am curious where/if components or assessment subjects fit into this layer. Would this be attached as an artifact or attribute of a system component defined elsewhere (e.g. in-toto Statement's subject)?

schemas/layer-4.cue
eddie-knight: I'm curious as to when we need layer 1 data this far up the stack 🤔 Especially since this is the _output_ from layer 4. 

I'd expect that both layer 5 and layer 6 will reference data from layer 3 and 4 directly, depending on the action — rather than trying to get policy data from the evaluation outputs.

eddie-knight: I think a different approach might be that an evaluation output could include an identifier for the policy it's measuring against. That policy would then have an ID to the guidelines and controls that they reference.

trumant:
> I am curious where/if components or assessment subjects fit into this layer. Would this be attached as an artifact or attribute of a system component defined elsewhere (e.g. in-toto Statement's subject)?

If I'm understanding correctly (or partially ;) ) @jpower432 - the language I use below might get at the same concepts:

Before I can successfully run a compliance program/initiative/what have you, I need to define a "compliance boundary" or "compliance scope". I have some inventory/knowledge of systems, components, configurations and processes and define whether each of them is in or out of that scope/boundary.

I could imagine that scope/boundary definition could live as Policy in Layer 3 perhaps? And if it did, then Layer 4 Evaluations would speak to applying controls (L2), across scope defined by (L3) and based on policy (L3).

Close or a million miles away from the thrust of your question?

trumant:
@jpower432 Some additional complexity that I foresee in Layer 4 is even when the scope/boundary is clearly defined the Evaluation also probably wants to communicate quite a bit about what data/facts are consumed from that scope, how trustworthy they are, and what the data quality is relative to the control requirement.

I threw together some quick thoughts recently on this in https://github.com/trumant/pvtr-github-repo/blob/redesign-payload-interface/architecture.md#input-evidence-data-and-sources

That was written in the context/from the perspective of `revanite-io/pvtr-github-repo` and therefore the compliance scope is effectively 1 single repository of source code. Your question hints at far larger scope to my eyes but I think the concerns can exist any scale.

I'm hopeful with feedback like yours we can continue to flesh out both Layer 4 and 3 to be more expressive and provide the "right level/granularity" of data for the consumers of it.

schemas/layer-4.cue
jpower432: @eddie-knight Your approach makes sense to me. I think a layer agnostic (allowing from Layer 2 or 3) control set identifier would allow for the same benefits I listed above, but I also see how referencing the policy information at Layer 6 could help to contextualize Layer 4 outputs.

eddie-knight:
Yes, my belief is that Layer 4 should be scoped to a single evaluation of a single resource. Although a set of evaluations could be triggered by a Layer 3 policy input.

jpower432:
> Before I can successfully run a compliance program/initiative/what have you, I need to define a "compliance boundary" or "compliance scope". I have some inventory/knowledge of systems, components, configurations and processes and define whether each of them is in or out of that scope/boundary.
> 

Thanks @trumant. That is exactly where I was going!
Generally, I was wondering how to define resources that fall within the scope of an assessment. I was also wondering if the category or "resource type" for the resource under scrutiny, along with associated data about the resource instance, would be explicitly captured in Layer 4.

> I could imagine that scope/boundary definition could live as Policy in Layer 3 perhaps? And if it did, then Layer 4 Evaluations would speak to applying controls (L2), across scope defined by (L3) and based on policy (L3).
 
Yes. This makes sense! I sounds like Layer 4 would not need "resource type" information explicitly because of the relationship with Layer 2 where that type would be inherent based on the defined technical boundaries.

schemas/layer-4.cue
```diff
@@ -1,37 +1,97 @@
...
+    // Remediation guide is a URL to remediation guidance associated with the control's assessment requirement and this specific assessment method.
+    remediation_guide?: #URL @go(RemediationGuide)
+    // Run is a boolean indicating whether the method was run or not. When run is true, result is expected to be present.
+    run: bool
 }
 
-#AssessmentResult: {
-    result: #Result
+// See https://cuelang.org/docs/tour/types/sumstruct/
+#AssessmentMethod: {
```
jpower432, line 53: Perhaps an optional `id` field could be added here. I think that could allow an external policy identifier to be used to perform the check in the event a direct check with the method executor is not used. Thoughts?

eddie-knight:
Now is a great time if we want to re-visit the concepts discussed here, but for this PR I suspect we'll have too much drift following other PRs that have been merged.

PR #20, opened 2025-04-22
wip: first attempt at a layer 3 schema

eddie-knight:
this is a conversation starter, and very far from being ready for actual review

### DIFF
--- a/schemas/layer-3.cue
+++ b/schemas/layer-3.cue
@@ -28,7 +28,7 @@ package layer3
     parent_policy_id?: string
 
     // Policy classification level (e.g., mandatory, recommended)
-    classification: oneOf("mandatory", "recommended")
+    classification: "mandatory" | "recommended"
 
     // Reference to one or more Layer 2 control catalogs
     control_catalogs: [...#CatalogReference]
--- a/schemas/layer-3.cue
+++ b/schemas/layer-3.cue
@@ -8,7 +8,8 @@ package layer3
     // Metadata useful for evaluation and automation
     metadata?: {
         version: string
-        author: string
+        // owner is the organizational unit responsible for the policy. This could be a department, team, or person.
+        owner: string @go(Owner)
         last_modified: string
         sci_version: string
         remarks: string
--- a/schemas/layer-3.cue
+++ b/schemas/layer-3.cue
@@ -10,8 +10,8 @@ import "time"
 #Policy: {
     // Metadata useful for evaluation and automation
     metadata?: {
-        // version if the version of the policy itself
-        version: string
+        // version is the version of the policy itself
+        version: string @go(Version)
         // owner is the organizational unit responsible for the policy. This could be a department, team, or person.
         owner: string @go(Owner)
         // last_modified is the date and time when the policy was last modified
@@ -50,15 +50,16 @@ import "time"
 
     // List of IDs of Layer 2 Catalog.Metadata.ApplicabilityCategories
     applicability: [...string] @Go(ApplicabilityCategories)
-
-    modify?: [#ControlModification] @go(Modify)
+    // one or more modifications to the controls in this catalog
+    modifications?: [#ControlModification, ...#ControlModification] @go(Modifications)
 
     // Reason for including this catalog in the policy
     objective?: string @go(Objective)
 }
 
 #ControlModification: {
-    id: string
+    // the ID of the control being modified
+    id: string @go(ID)
 
     // The modified applicability level of this control, using IDs defined in the catalog
     // An empty list means the control should be omitted entirely
@@ -66,10 +67,4 @@ import "time"
 
     // Justification for modifying this control
     rationale?: string @go(Rationale)
-}
-
-#Reference: {
-    title: string @go(Title)
-    url?: string @go(URL)
-    description?: string @go(Description)
 }
--- a/schemas/layer-3.cue
+++ b/schemas/layer-3.cue
@@ -1,5 +1,5 @@
 // layer-3-policy.cue
-
+package layer3
 // TODO
 // Create schema for a policy document
 // Policy docs should reference common and unique risk definitions, so we need schema for that as well
--- a/schemas/layer-3.cue
+++ b/schemas/layer-3.cue
@@ -1,0 +1,67 @@
+// layer-3-policy.cue
+
+// TODO
+// Create schema for a policy document
+// Policy docs should reference common and unique risk definitions, so we need schema for that as well
+
+#Policy: {
+    // Metadata useful for evaluation and automation
+    metadata?: {
+        version: string
+        author: string
+        last_modified: string
+        sci_version: string
+        remarks: string
+    }
+
+    // Unique identifier for this policy
+    id: string
+
+    // Human-readable title of the policy
+    title: string
+
+    // Short description of this policy’s intent or purpose
+    description: string
+
+    // Optional reference to a parent policy this one inherits from or refines
+    parent_policy_id?: string
+
+    // Policy classification level (e.g., mandatory, recommended)
+    classification: oneOf("mandatory", "recommended")
+
+    // Reference to one or more Layer 2 control catalogs
+    control_catalogs: [...#CatalogReference]
+
+}
+
+#CatalogReference: {
+
+    id: string
+
+    version: string
+
+    // List of IDs to applicability values defined in the catalog
+    applicability: [...string]
+
+    modify?: [#ControlModification]
+
+    // Reason for including this catalog in the policy
+    objective?: string
+}
+
+#ControlModification: {
+    id: string
+
+    // The modified applicability level of this control, using IDs defined in the catalog
+    // An empty list means the control should be omitted entirely
+    applicability: [...string]
+
+    // Justification for modifying this control
+    rationale?: string
+}
+
+#Reference: {
+    title: string
+    url?: string
+    description?: string
+}
--- a/schemas/layer-3.cue
+++ b/schemas/layer-3.cue
@@ -31,7 +31,7 @@ package layer3
     classification: "mandatory" | "recommended"
 
     // Reference to one or more Layer 2 control catalogs
-    control_catalogs: [...#CatalogReference]
+    control_catalogs: [#CatalogReference, ...#CatalogReference]
 
 }
 
--- a/schemas/layer-3.cue
+++ b/schemas/layer-3.cue
@@ -1,5 +1,8 @@
 // layer-3-policy.cue
 package layer3
+
+import "time"
+
 // TODO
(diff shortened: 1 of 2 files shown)


schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+#CatalogReference: {
+
+    id: string
+
+    version: string
+
+    // List of IDs to applicability values defined in the catalog
+    applicability: [...string]
+
+    modify?: [#ControlModification]
```
trumant:
```suggestion
    modify?: [#ControlModification...]
``` 

_maybe....._

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
+// layer-3-policy.cue
+
```
trumant, line 5:
```suggestion
package layer3
```

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+// TODO
+// Create schema for a policy document
+// Policy docs should reference common and unique risk definitions, so we need schema for that as well
+
```
trumant, line 9: Providing/linking to some example policies that you imagine encoding/decoding with the schema would be helpful

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+#Policy: {
+    // Metadata useful for evaluation and automation
+    metadata?: {
+        version: string
+        author: string
```
trumant:
```suggestion
```
recommend dropping until we better understand or explain the need. I suspect most policies are authored by multiple individuals and I'm also not sure to what extent its important to identify the author(s) for consumers of data conforming to the schema.

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
+// layer-3-policy.cue
+
+// TODO
+// Create schema for a policy document
+// Policy docs should reference common and unique risk definitions, so we need schema for that as well
+
+#Policy: {
+    // Metadata useful for evaluation and automation
+    metadata?: {
+        version: string
```
trumant: Opportunity to define/reference a SemVer or similar definition rather than simply string, but string seems like a decent starting point.

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+        author: string
+        last_modified: string
```
trumant: What's the datetime format, we should specify it in the type constraint, but really just reference an existing definition from elsewhere

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+        sci_version: string
```
trumant: https://github.com/revanite-io/sci/issues/25

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+    id: string
+
+    // Human-readable title of the policy
+    title: string
+
+    // Short description of this policy’s intent or purpose
+    description: string
+
+    // Optional reference to a parent policy this one inherits from or refines
+    parent_policy_id?: string
```
trumant: Feels like the tip of a complexity iceberg, but likely a very valid idea we could choose to model.

I'd like to see more discussion how this is interpreted before reaching any definitive conclusions on its necessity.

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+
+    // Policy classification level (e.g., mandatory, recommended)
+    classification: oneOf("mandatory", "recommended")
```
trumant: I don't see `oneOf` in https://cuelang.org/docs/reference/spec/#builtin-functions

Should this be
```suggestion
    classification: "mandatory" | "recommended"
```

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+
+    // Reference to one or more Layer 2 control catalogs
+    control_catalogs: [...#CatalogReference]
```
trumant: As written, the cue constraint is 0 or more `#CatalogReference`s
```suggestion
    control_catalogs: [#CatalogReference, ...#CatalogReference]
```

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+
+}
+
+#CatalogReference: {
+
+    id: string
+
+    version: string
+
+    // List of IDs to applicability values defined in the catalog
```
trumant: I think this points to a need to better document/define applicability as a concept. I'm not following

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+    applicability: [...string]
+
```
trumant:
```suggestion

// is the referenced control from the catalog reference being modified by policy?
```

I think the entire concept of policy modification of controls needs some useful examples to ensure shared understanding.

I think that it relates to cases where a control has language that can incorporate organization-defined policy as "user-supplied variables" for example NIST 800-53r5 control AC-02(03):

### Control language

> Disable accounts within {{ insert: param, ac-02.03_odp.01 }} when the accounts:
> 
> (a): Have expired;
> (b): Are no longer associated with a user or individual;
> (c): Are in violation of organizational policy; or
> (d): Have been inactive for {{ insert: param, ac-02.03_odp.02 }}.

> ID: ac-02.03_odp.01
> Label: time period
> Guidelines:
> time period within which to disable accounts is defined;
> ID: ac-02.03_odp.02
> Label: time period
> Guidelines:
> time period for account inactivity before disabling is defined;

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+    // The modified applicability level of this control, using IDs defined in the catalog
+    // An empty list means the control should be omitted entirely
+    applicability: [...string]
+
+    // Justification for modifying this control
+    rationale?: string
+}
+
+#Reference: {
```
trumant: I don't see this currently used - how are you imagining it fits in with the other defs?

schemas/layer-3.cue
eddie-knight: Looking again, this should be something like "owner"

schemas/layer-3.cue
```diff
@@ -0,0 +1,67 @@
...
+package layer3
+// TODO
+// Create schema for a policy document
+// Policy docs should reference common and unique risk definitions, so we need schema for that as well
+
+#Policy: {
+    // Metadata useful for evaluation and automation
+    metadata?: {
+        version: string
+        author: string
```
trumant:
```suggestion
        // owner is the organizational unit responsible for the policy. This could be a department, team, or person.
        owner: string @go(Owner)
```

schemas/layer-3.cue
trumant: The `pvtr-github-repo` configuration has a notion of policy which seems to have been defined by `privateer-sdk`

Example configuration I'm referring to is:

```yaml
# The policy section may be set at the top level if assessing multiple services via privateer
    policy:
      catalogs:
        - OSPS_B # currently this is the only available catalog to assess against
      applicability:
        - Maturity Level 1
        # - Maturity Level 2
        # - Maturity Level 3
```

and the relevant code in the sdk is: https://github.com/privateerproj/privateer-sdk/blob/main/pluginkit/vessel.go#L77

schemas/layer-3.cue
eddie-knight: might get away with just using `time.Time` here for now.

schemas/layer-3.cue
eddie-knight: I suspect this is leftover from something else & can be removed

schemas/layer-3.cue
trumant: @eddie-knight As I dove further into my branch with generated types and tried to implement Evaluate, I ran into a perfect use case for Layer 3 to bridge between 2 and 4, see https://github.com/trumant/sci/pull/1/files#diff-9902cbfa7faa6b2d39761d5085851aa84657f1acf16dcc400480ba4bcbaa59d4R14-R22

Curious to hear your view on where you see the filtering occurring

trumant: https://github.com/trumant/sci/pull/1/files#diff-f516c03e4831eac9448107db61b07347cdba8023233c85638cb207faadded288 also shows where policy could play a role

trumant (commented):
After we chatted @eddie-knight , I gave this a fresh review and took the following notes. These are presented in order of the observations/conclusions I reached as I read through the schema file in-toto absent its current review commentary.

### Notes

#### Cool pattern, please extend for consistency

I love the introduction of SCI metadata and layer metadata. Could you perhaps to start to flesh this out further by introducing a `shared.cue` or similar and define a type like the following that each layer can then use?

I.e. for these particular fields, if we introduce them in l3, it would be odd for them to be missing from l2 and l4

```cue
metadata {
    last_modified: 2023-04-28
    sci_version: 0.1.0
}
```

#### parent_policy_id

- Can you elaborate on how this value (along with others) would express inheritance, refinement and composition?
- Does this imply unbounded lineage? If so, can we choose differently and bound it somehow?
- Is this needed right out of the gate? Put another way, if we didn't model it and implement inheritance, refinement and composition, who would suffer and how?

#### classification

I love a good enum, but I have questions
- Why should an author of a policy choose one or the other value?
- How are the values meant to be interpreted by systems consuming layer 3 and producing layer 4?

#### CatalogReference

Given an example ID and version value, how would a consumer be able to source and load the catalog? Does this type need a URL field to communicate that location?

#### As a system consuming layer 3 to execute checks and produce layer 4

The policy evaluation might look something like:

Given a policy
  Check for parent_policy_id and if present
    Fetch that policy and recurse until parent_policy_id is nil
    perform some aggregate evaluation of the policy lineage to produce a single policy, final_policy

Given final_policy
    Check classification and do something based on its value
        Unclear what that something is
    
Iterate over each CatalogReference
  Source and load the Catalog based on id and version

  Evaluate the applicability and modifications values to draw some conclusion about which controls in the catalog are required by policy to produce a filtered selection of all controls in the catalog

schemas/layer-3.cue
```diff
@@ -0,0 +1,70 @@
...
+    // List of IDs of Layer 2 Catalog.Metadata.ApplicabilityCategories
+    applicability: [...string] @Go(ApplicabilityCategories)
+    // one or more modifications to the controls in this catalog
+    modifications?: [#ControlModification, ...#ControlModification] @go(Modifications)
+
+    // Reason for including this catalog in the policy
+    objective?: string @go(Objective)
+}
+
+#ControlModification: {
```
jpower432, line 60: It would likely require an update to Layer 2, but I think organizational defined parameters could be useful as part of the control modification. Layer 2 could define possible parameters with associated constraints, perhaps even customizing by level.

PR #28, opened 2025-04-23
test: add test coverage for layer2/loaders

trumant:


pkg/layer2/loaders.go
```diff
@@ -101,3 +99,13 @@ func (c *Catalog) LoadFile(sourcePath string) error {
...
 }
+
+func decode(reader io.Reader, data *Catalog) error {
+	decoder := yaml.NewDecoder(reader)
+	decoder.KnownFields(true)
+	err := decoder.Decode(data)
+	if err != nil {
+		return fmt.Errorf("error decoding YAML: %w", err)
+	}
+	return nil
```
eddie-knight, line 110:
```suggestion
	return decoder.Decode(data)
```

nit: Since we're wrapping this with a message based on whether it's from url or file, we probably don't need to wrap it here as well

eddie-knight (approved):

PR #43, opened 2025-04-26
chore: improved logging for layer2.Load

eddie-knight:
As a developer user, I often load multiple files at once using SCI. When I get a load error, I would like to know which file caused the error.


trumant (approved):
Right on!

eddie-knight:
shucks, merged the PRs out of order. new commit incoming.

PR #44, opened 2025-04-26
chore: moved modules from pkg to root dir

### LINKED ISSUE #31: [go][docs] remove `pkg` directory from the module source
# Overview

Today the go source for layers 2 and 4 are in `./pkg/layer2` and `./pkg/layer4` which does not follow the recommendations in https://go.dev/doc/modules/managing-source#repository which results in a slightly awkward experience for consumers of the module when managing imports and navigating https://pkg.go.dev/github.com/revanite-io/sci

# Dependencies

1. For ease of adoption by existing consumers, this change should be made at the same time as the work done to migrate the module into its new Github org assuming that consensus/acceptance is reached on #24

eddie-knight:
Tested this change locally in a build of [controls-canvas](https://github.com/revanite-io/controls-canvas); everything behaves as expected with no change aside from removing pkg from the import path.

Resolves #31


trumant (approved):
Bravo!

Issue #35, opened 2025-04-24
[schema] standardize on either "name" or "title" field names, but stop using both randomly


trumant:
# Overview

A good schema is one that is internally consistent, because internal consistency allows schema consumers to anticipate the likely shape of types based on past experience with other types in the same schema.

# Dependencies

1. As this constitutes a breaking change for existing consumers, this change should be made at the same time as the work done to migrate the module into its new Github org assuming that consensus/acceptance is reached on https://github.com/revanite-io/sci/issues/24

# Next Steps

- [x] Pick a winner term, personally I'm partial to `name` over `title`
- [x] Do the needful across the relevant schema definitions

eddie-knight:
Agree on the need for consistency.

Looking at layer2 right now, especially [here](https://github.com/finos/common-cloud-controls/tree/main/services/storage/object), and my gut says `name` just doesn't fit for the cases where we're using `title`.

eddie-knight:
I've created PR #45 to address some erroneous blending in the layer2 go module, as the cue schema is consistent on using `title` with no instances of `name`.

The intent here is to use titles in situations where a unique identifier is present. The titles supplement the unique `id` value with unique, semi-static, but non-functional text.

Inversely, we have `name` without `title` in layer4.

The intent of a `name` value is to provide instance-specific identifying data. Where titles are conceptual, names are tangible. 

Upon inspection, I'm noticing that we don't always use (or need) these names. For example, `pvtr-github-repo` is not currently filling in names for `control_evaluation` entries.

My next step will be to examine where the name fields are being used in a way that matches the intent while providing end-user value. After I poke at it for a bit, I'll propose a change accordingly.

jpower432:
We completed this work

PR #45, opened 2025-04-26
fix: Match layer2 module to layer2 schema (name/title)

eddie-knight:
follows PR #44 

relates to #35 

fixes the drift between the layer2 go schema and the layer2 cue schema

Tested with a local version of [controls-canvas](https://github.com/revanite-io/controls-canvas) after changing to use local CCC yaml files with control family name->title. Results were as expected.


Makefile
```diff
@@ -10,13 +10,13 @@ release: tidy test release-nix release-win release-mac
 
 define build
 echo "  >  Building binary ..."; \
-cd pkg/$(1); \
+cd $(1); \
 go build -o ../../$(1) -ldflags="$(BUILD_FLAGS)";
```
trumant:
```suggestion
go build -o ../dist/$(1) -ldflags="$(BUILD_FLAGS)";
```
^ that along with `mkdir dist && touch dist/.gitkeep` perhaps

trumant (approved):
This makes sense to me. I had a small suggestion unrelated to field naming for the `Makefile`, but that's definitely non-blocking feedback.

Makefile
eddie-knight: mind if we do that in a follow-up? Seems unrelated to the current change set

PR #51, opened 2025-04-28
feat: one schema with objects for each layer

eddie-knight:
Combining the schemas into one, so we can share objects if needed

### DIFF
--- a/schemas/layer-2.cue
+++ b/schemas/layer-2.cue
@@ -1,7 +1,6 @@
-package layer2
-// Top level schema //
+package schemas
 
-catalog: {
+#Layer2: {
     metadata?: #Metadata
 
     "control-families"?: [...#ControlFamily]
--- a/schemas/layer-4.cue
+++ b/schemas/layer-4.cue
@@ -1,10 +1,8 @@
-package layer4
-// Top level schema //
+package schemas
 
-#EvaluationPlan: {
+#Layer4: {
+    evaluations: [#ControlEvaluation, ...#ControlEvaluation]
 }
-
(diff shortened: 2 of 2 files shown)


trumant (approved):

eddie-knight:
@trumant Should we call this `Evaluation` and `#Catalog` (or `#ControlCatalog`) instead?

Part of me is thinking that we shouldn't hardcode the "layer" language if possible, but I don't have a good justification to tip the scales either direction.

trumant:
> @trumant Should we call this `Evaluation` and `#Catalog` (or `#ControlCatalog`) instead?
> 
> 
> 
> Part of me is thinking that we shouldn't hardcode the "layer" language if possible, but I don't have a good justification to tip the scales either direction.

The names are definition names so schema specific. The data can use whatever key it wants and still unify against the definition.

Put another way the yaml key in the data can be catalog and still be validated against a definition with whatever name.

But, it's more obvious and expected that the names match for both users and maintainers. And because we expect to generate Go and hopefully also Markdown based on the Cue content we should follow the [Principle of Least Surprise/Astonishment](https://en.m.wikipedia.org/wiki/Principle_of_least_astonishment) I think.

Put another way, do we want the cue definition names, the YAML data to use the layer4 and layer2 keys and for the go structs to be named Layer2 and Layer4 - that seems advantageous to overall understanding.

If you see this change flowing through all of those layers and reducing astonishment than I'm all for it.

eddie-knight:
Thanks for the feedback, I concur.

The entire point of the "layer" language is to help users have flawless clarity on when and where an activity and schema belong.

PR #54, opened 2025-05-06
docs: add comments to layer-2 schema for better docs

trumant:
This PR adds comments for nearly all definitions and fields in `schemas/layer-2.cue` in order to better explain the schema. As the go types are now generated from the schema, this has the happy side effect of improving the go module docs as well.

### DIFF
--- a/layer2/test-data/good-ccc.yaml
+++ b/layer2/test-data/good-ccc.yaml
@@ -3,7 +3,7 @@ metadata:
   description: |
     FINOS CCC is an open standard project that describes consistent controls for
     compliant public cloud deployments in the financial services sector.
-  applicability-categories:
+  assessment-levels:
     - id: tlp_clear
       title: TLP:Clear
       description: |
@@ -38,7 +38,7 @@ control-families:
           - reference-id: CCC
             identifiers:
               - CCC.TH02 # Data is Intercepted in Transit
-        guideline-mappings:
+        guidance-mappings:
           - reference-id: CSF
             identifiers:
               - PR.DS-02 # Data-in-transit is protected
@@ -58,7 +58,7 @@ control-families:
             text: |
               When a port is exposed for non-SSH network traffic, all traffic MUST
               include a TLS handshake AND be encrypted using TLS 1.2 or higher.
-            applicability:
+            levels:
               - tlp_clear
               - tlp_green
               - tlp_amber
@@ -67,7 +67,7 @@ control-families:
             text: |
               When a port is exposed for SSH network traffic, all traffic MUST
               include a SSH handshake AND be encrypted using SSHv2 or higher.
-            applicability:
+            levels:
               - tlp_clear
               - tlp_green
               - tlp_amber
@@ -85,7 +85,7 @@ control-families:
           - reference-id: CCC
             identifiers:
               - CCC.TH03 # Deployment Region Network is Untrusted
-        guideline-mappings:
+        guidance-mappings:
           - reference-id: CCM
             identifiers:
               - DSI-06
@@ -105,7 +105,7 @@ control-families:
               When a deployment request is made, the service MUST validate
               that the deployment region is not to a restricted or regions
               or availability zones.
-            applicability:
+            levels:
               - tlp_clear
               - tlp_green
               - tlp_amber
@@ -115,7 +115,7 @@ control-families:
               When a deployment request is made, the service MUST validate that
               replication of data, backups, and disaster recovery operations
               will not occur in restricted regions or availability zones.
-            applicability:
+            levels:
               - tlp_clear
               - tlp_green
               - tlp_amber
@@ -131,7 +131,7 @@ control-families:
           - reference-id: CCC
             identifiers:
               - CCC.TH06 # Data is Lost or Corrupted
-        guideline-mappings:
+        guidance-mappings:
           - reference-id: CSF
             identifiers:
               - PR.DS-5 # Protections against data leaks are implemented
@@ -147,7 +147,7 @@ control-families:
             text: |
               When data is stored, the service MUST ensure that data is
               replicated across multiple availability zones or regions.
-            applicability:
+            levels:
               - tlp_green
               - tlp_amber
               - tlp_red
@@ -157,7 +157,7 @@ control-families:
               the service MUST be able to verify the replication state,
               including the replication locations and data synchronization
               status.
-            applicability:
+            levels:
               - tlp_green
               - tlp_amber
               - tlp_red
@@ -174,7 +174,7 @@ control-families:
               - CCC.TH07 # Logs are Tampered with or Deleted
               - CCC.TH09 # Logs or Monitoring Data are Read by Unauthorized Users
               - CCC.TH04 # Data is Replicated to Untrusted or External Locations
-        guideline-mappings:
+        guidance-mappings:
           - reference-id: CCM
             identifiers:
               - LOG-02 # Audit log protection
@@ -188,7 +188,7 @@ control-families:
             text: |
               When access logs are stored, the service MUST ensure that
               access logs cannot be accessed without proper authorization.
-            applicability:
+            levels:
               - tlp_amber
               - tlp_red
               - tlp_green
@@ -197,7 +197,7 @@ control-families:
             text: |
               When access logs are stored, the service MUST ensure that
               access logs cannot be modified without proper authorization.
-            applicability:
+            levels:
               - tlp_amber
               - tlp_red
               - tlp_green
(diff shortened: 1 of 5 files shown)


schemas/layer-2.cue
```diff
@@ -1,84 +1,131 @@
...
     title: string
+    // description of the Layer2 collection
     description: string
     version?: string
+    // timestamp of the last modification to the Layer2 collection
     "last-modified"?: string @go(LastModified)
+    // Applicability categories are facets that apply to the controls in this Layer2 collection. These facet values can be used to filter which controls from Layer2 are used in Layer 3 or Layer 4.
     "applicability-categories"?: [...#Category] @go(ApplicabilityCategories)
+    // List of applicable references to Layer 1 guidance, technical capabilities and threats that inform the Layer2 collection
     "mapping-references"?: [...#MappingReference] @go(MappingReferences)
```
trumant, line 35: See https://github.com/ossf/security-baseline/blob/main/baseline/frameworks.yaml for an example

schemas/layer-2.cue
```diff
@@ -1,84 +1,131 @@
...
     id: string
+    // name for the Layer2 collection
     title: string
+    // description of the Layer2 collection
     description: string
     version?: string
+    // timestamp of the last modification to the Layer2 collection
     "last-modified"?: string @go(LastModified)
+    // Applicability categories are facets that apply to the controls in this Layer2 collection. These facet values can be used to filter which controls from Layer2 are used in Layer 3 or Layer 4.
     "applicability-categories"?: [...#Category] @go(ApplicabilityCategories)
```
trumant: For the OSPS Baseline, I _think_ this could look something like:

```yaml
applicability-categories:
  - id: level1
    name: Level 1
    description: for any code or non-code project with any number of maintainers or users
  - id: level2
    name: Level 2
    description: for any code project that has at least 2 maintainers and a small number of consistent users
  - id: level3
    name: Level 3
    description: for any code project that has a large number of consistent users
```

schemas/layer-2.cue
```diff
@@ -1,84 +1,131 @@
...
 #Metadata: {
+    // unique identifier for the Layer2 collection
     id: string
+    // name for the Layer2 collection
     title: string
+    // description of the Layer2 collection
     description: string
     version?: string
+    // timestamp of the last modification to the Layer2 collection
     "last-modified"?: string @go(LastModified)
```
trumant, line 31: Should we take this opportunity to make this `time.Time`?

schemas/layer-2.cue
```diff
@@ -1,84 +1,131 @@
...
+    // name of the control
     title: string
+    // the intended outcome of applying the control
     objective: string
+    // the assessment requirements that are used to determine if the control is met
     "assessment-requirements": [...#AssessmentRequirement] @go(AssessmentRequirements)
-
-    "guideline-mappings"?: [...#Mapping] @go(GuidelineMappings)
+    // references to layer 1 guidance or standards that inform the control
+    "guidance-mappings"?: [...#Mapping] @go(GuidanceMappings)
```
trumant, line 69: Note the rename here and the fact that baseline will need an update either way, see https://github.com/ossf/security-baseline/blob/main/baseline/OSPS-LE.yaml#L22

schemas/layer-2.cue
```diff
@@ -1,84 +1,131 @@
...
     version?: string
+    // timestamp of the last modification to the Layer2 collection
     "last-modified"?: string @go(LastModified)
+    // Applicability categories are facets that apply to the controls in this Layer2 collection. These facet values can be used to filter which controls from Layer2 are used in Layer 3 or Layer 4.
     "applicability-categories"?: [...#Category] @go(ApplicabilityCategories)
+    // List of applicable references to Layer 1 guidance, technical capabilities and threats that inform the Layer2 collection
     "mapping-references"?: [...#MappingReference] @go(MappingReferences)
 }
 
+// Category defines a facet that applies to the AssessmentRequirements in Controls. Categories typically define "levels" of compliance, for example the low, moderate and high baselines within the NIST 800-53 controls.
```
funnelfiasco: Is there a reason we can't use "level" here? Conceptually, "category" implies a vertically-stacked set of entries on a similar theme, whereas "level" is a horizontal slice across all categories.

Using OSPS Baseline as an example, the families like "Legal" and "Documentation" are categories and levels are..."levels" 😄 

Or to use an org chart analogy, "software engineer", "project manager", and "technical marketing manager" are categories, while "junior", "senior", and "principal" are levels.

Does that make sense?

schemas/layer-2.cue
eddie-knight: Let's not add more items to the cognitive load for this review

schemas/layer-2.cue
eddie-knight: Yep. or snake/kebab, whatever.

schemas/layer-2.cue
eddie-knight: ++ This should also allow Baseline to avoid putting guidance links in the Lexicon as those could be auto-generated from this data

schemas/layer-2.cue
trumant: Or perhaps rename to "facet" and let either a category name or level name be used?

While nearly all compliance frameworks I've seen use levels, I suspect they aren't the only way of slicing and dicing a catalog of controls.

Maybe a bigger question - do we want the schema to be more restrictive or flexible in this regard? Once we have a sense of that, perhaps that clarifies naming.

SecurityCRob (approved):

schemas/layer-2.cue
funnelfiasco: "Facet" isn't immediately intuitive to me. For my particular consumption (OSPS Baseline), "level" makes sense, but I don't want to dictate that everyone be like us.

That said, if most/all frameworks use something that can reasonably be called levels, I think it's worth being a little restrictive here. There are hypothetically other ways to carve up the catalog, but too much flexibility sacrifices clarity. If there are notable examples of using something other than levels, I'm open to finding a broader term.

eddie-knight: I'm not reflexively opposed to levels as a term here.

I think most of the thoughts leading away from that term were attempting to avoid ladder-style verbiage, but... meh. Everything in this space can be sorted onto a ladder in some way. Level is fine.

trumant: Appreciate the feedback and discussion here. I've pushed changes that try to address this naming feedback.

layer2/test-data/good-ccc.yaml
```diff
@@ -3,7 +3,7 @@ metadata:
   description: |
     FINOS CCC is an open standard project that describes consistent controls for
     compliant public cloud deployments in the financial services sector.
-  applicability-categories:
+  assessment-levels:
```
eddie-knight, line 6: Oh interesting... this isn't a wild approach, but it is unexpected.

I was thinking we'd have `assessment.applicability` as a list of `applicability-levels` ids. So the only change would be here, and everything lower would be the same. (rather than changing to `assessment.levels`)

Do you think this larger change is worthwhile?

schemas/layer-2.cue
```diff
@@ -1,84 +1,133 @@
...
     id: string
+    // name for the Layer2 collection
     title: string
+    // description of the Layer2 collection
     description: string
     version?: string
+    // timestamp of the last modification to the Layer2 collection
     "last-modified"?: string @go(LastModified)
-    "applicability-categories"?: [...#Category] @go(ApplicabilityCategories)
+    // AssessmentLevels is a list of values used to categorize the AssessmentRequirements of the Controls in this Layer2 collection.For example, the NIST 800-53 controls are categorized as low, moderate, and high baselines and if this Layer2 collection contained those NIST 800-53 controls, the AssessmentLevels would be "low", "moderate", and "high".
```
trumant:
```suggestion
    // AssessmentLevels is a list of values used to categorize the AssessmentRequirements of the Controls in this Layer2 collection. For example, the NIST 800-53 controls are categorized as low, moderate, and high baselines and if this Layer2 collection contained those NIST 800-53 controls, the AssessmentLevels would be "low", "moderate", and "high".
```

layer2/test-data/good-ccc.yaml
trumant: I think its extremely valuable for the naming to be consistent between the two fields as it will lead to better understanding.

schemas/layer-2.cue
```diff
@@ -1,84 +1,132 @@
 package schemas
 @go(layer2)
 
+// Layer2 describes a collection or catalog of technology-specific, threat-informed security controls
```
eddie-knight:
```suggestion
// Catalog describes a collection or catalog of technology-specific, threat-informed security controls
```

eddie-knight:
This conversation sprawled a bit and I'm not sure it'll wrap up on the current thread. Lots of good points here, but I'm going to close this PR — let's reopen with single-scope changes as needed.

PR #56, opened 2025-05-08
ci: lint our security-insights.yml in CI workflow

trumant:
Inspired by the work in https://github.com/revanite-io/sci/pull/55

I realized as I was reviewing that PR, that I really want the CI workflow to give me basic validation feedback before spending time on my review.

Which led me here, only to find and then fix a bunch of validation issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,4 +86,11 @@ dirtycheck:
 		echo "  >  No uncommitted changes to generated files found."; \
 	fi
 
+lintinsights:
+	@echo "  >  Linting security-insights.yml ..."
+	@curl -O --silent https://raw.githubusercontent.com/ossf/security-insights-spec/refs/tags/v2.1.0/schema.cue
+	@cue vet security-insights.yml schema.cue
+	@rm schema.cue
+	@echo "  >  Linting security-insights.yml complete."
+
 PHONY: lintcue cuegen dirtycheck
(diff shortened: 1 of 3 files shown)


security-insights.yml
```diff
@@ -14,13 +14,17 @@ project:
     - name: Jason Meridth
       affiliation: GitHub
       email: jmeridth@gmail.com
+      primary: false
```
eddie-knight, line 17: Interesting place to be explicit, but I'm fine with it.

Makefile
```diff
@@ -86,4 +86,11 @@ dirtycheck:
 		echo "  >  No uncommitted changes to generated files found."; \
 	fi
 
+lintinsights:
+	@echo "  >  Linting security-insights.yml ..."
+	@curl -O --silent https://raw.githubusercontent.com/ossf/security-insights-spec/refs/tags/v2.0.0/schema.cue
+	@cue vet security-insights.yml schema.cue
+	@rm schema.cue
+	@echo "  >  Linting security-insights.yml complete."
+
```
eddie-knight: this could be the world's simplest GitHub Action, if we wanted to enable SI vetting for any/everyone 🤔 someday

jpower432: +1 to a GitHub Action. I would be interested in using that!

eddie-knight (approved):

PR #63, opened 2025-05-16
ci: add a composite GH Action for SI linting

### LINKED ISSUE #62: Create a GitHub Action to vet against schema
> this could be the world's simplest GitHub Action, if we wanted to enable SI vetting for any/everyone 🤔 someday

_Originally posted by @eddie-knight in https://github.com/revanite-io/sci/pull/56#discussion_r2080839863_

jpower432:
Add a top level `actions` directory with a composite action for linting the security insights file.

A couple points which could be up for discussion:
- Adds cue setup as an opt-out step because I figured those consuming this likely will not have this dependency installed
- `actions` is in the root of the repository rather that `.github/actions` for discoverabilty 

Closes #62

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Validate the security-insights file
         uses: ./actions/lint-insights
         with:
-          cue_setup: 'false'
+          cue_setup: "false"
       - name: Generate go types from cue schema
         run: make cuegen
       - name: Check for missing generated artifacts
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Validate the layer 2 schema
         run: make lintcue
       - name: Validate the security-insights file
-        uses: ./actions/lint-insights
+        uses: ./.github/actions/lint-insights
         with:
           cue_setup: "false"
       - name: Generate go types from cue schema
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,9 @@ jobs:
       - name: Validate the layer 2 schema
         run: make lintcue
       - name: Validate the security-insights file
-        run: make lintinsights
+        uses: ./actions/lint-insights
+        with:
+          cue_setup: 'false'
       - name: Generate go types from cue schema
         run: make cuegen
       - name: Check for missing generated artifacts
--- a/actions/lint-insights/action.yaml
+++ b/actions/lint-insights/action.yaml
@@ -1,0 +1,36 @@
+name: 'Lint Security Insights File'
+description: 'Use cue to validate a given security insights file against the SI schema.'
+
+inputs:
+  si_path:
+    description: 'Relative security-insights.yml path'
+    required: false
+    default: 'security-insights.yml'
+  schema_version:
+    description: 'Version of the SI schema to use'
+    required: false
+    default: 'v2.1.0'
+  cue_setup:
+    description: 'Run setup-cue (defaults to true)'
+    required: false
+    default: true
+
+runs:
+  using: "composite"
+  steps:
+    - if: inputs.cue_setup == 'true'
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -22,7 +22,9 @@ jobs:
       - name: Validate the layer 2 schema
         run: make lintcue
       - name: Validate the security-insights file
-        run: make lintinsights
+        uses: ./actions/lint-insights
+        with:
+          cue_setup: 'false'
```
trumant: Nit: consistent string quoting

```suggestion
          cue_setup: "false"
```

actions/lint-insights/action.yaml
```diff
@@ -0,0 +1,36 @@
+name: 'Lint Security Insights File'
```
trumant: Nit: https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action#creating-a-composite-action-within-the-same-repository suggests that composite actions used within the repository should live in `.github/actions`

actions/lint-insights/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+    required: false
+    default: 'security-insights.yml'
+  schema_version:
+    description: 'Version of the SI schema to use'
+    required: false
+    default: 'v2.1.0'
+  cue_setup:
+    description: 'Run setup-cue (defaults to true)'
+    required: false
+    default: true
```
trumant: https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#inputsinput_iddefault says this is a string, so let's quote it.

```suggestion
    default: "true"
```

actions/lint-insights/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+      uses: cue-lang/setup-cue@a93fa358375740cd8b0078f76355512b9208acb1 # v1.0.1
+      with:
+        version: "v0.13.0-alpha.4"
+
+    - name: Lint security-insights.yml
+      run: |
+        echo "::group::Linting security-insights.yml ..."
+        curl -O --silent "https://raw.githubusercontent.com/ossf/security-insights-spec/refs/tags/${VERSION}/schema.cue"
+        cue vet -d '#SecurityInsights' "${SI_PATH}" schema.cue
+        rm schema.cue
```
trumant: My strong preference is to not duplicate this logic and instead find a way that this can simply run `make lintinsights`. My rationale is to ensure that a contributor can get be sure they will get the same result locally as they are going to get in CI.

trumant (changes requested):

trumant (commented):
I ran `zizmor actions` against this branch and was delighted to see:

```
INFO audit: zizmor: 🌈 completed actions/lint-insights/action.yaml
No findings to report. Good job!
```

actions/lint-insights/action.yaml
jpower432: Thanks for the feedback! Given that environment variables are being used here, I think I can add a simple update to the Makefile and call it in the action  :+1: .

actions/lint-insights/action.yaml
jpower432: You bring up a great point. From the original comment, I was thinking the intent is that the action be imported in CI workflows in other repositories. However, that probably makes more sense if it was located in the `ossf/security-insights-spec` or `ossf/si-tooling` repository.

jmeridth:
Could we make this live in its own repository and an actual action or is this to test it? By making it its own repo and a standalone action, others can use more easily. access it via the marketplace and its cadence for work is isolated. Or should it live in https://github.com/ossf/si-tooling (I prefer standalone still). Being in this repo feels a little weird. wdyt?

eddie-knight:
I've raised a similar question on #62 @jmeridth

jpower432:
I put this in draft while we continue the discussion on #62 on the proper location for this

jpower432:
Closing this PR and moving action to https://github.com/revanite-io/security-insights-linter

PR #66, opened 2025-06-09
fix(metadata): Updating Metadata for YAML Unmarshalling

damienjburks:
Fixing this exception we're getting for the CCC Delivery pipeline:

```bash
line 9: field last-modified not found in type layer2.Metadata
line 10: field applicability-categories not found in type layer2.Metadata
line 11: field mapping-references not found in type layer2.Metadata
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,21 @@ cuegen:
 	@echo "  >  Generating types for layer2 ..."
 	@cue exp gengotypes ./schemas/layer-2.cue
 	@mv cue_types_gen.go layer2/generated_types.go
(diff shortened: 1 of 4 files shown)


trumant (commented):
I'm not sure I fully understand the intent of these changes. Is there an issue or more details you can provide around what behavior you are trying to change and why?

eddie-knight:
Just created #67 with my understanding of this issue.

Ideally we would be able to use the generated go types for everything, but Damien reports finding that the bug report is mitigated by adding yaml tags to the necessary fields.

PR #76, opened 2025-08-06
fix: adds yaml struct tags to layer1 generated types

eddie-knight:
This will be an unverified commit because it's a cherry-pick from the divergent Gemara repo. Thanks/apologies @jpower432


trumant:
Closing in favor of #78

PR #77, opened 2025-08-06
feat: Added layer3 schema and go package

eddie-knight:
This is a smashup of old commits I made related to Layer 3. Some timestamps are lost, but that's it.


jpower432 (approved):
LGTM

Issue #23, opened 2025-04-22
[design] where do attestations and evidence fit into the layers?


trumant:
# Overview

Layer 4 assessments evaluate input data which could include attestations of compliance to one or more controls and their requirements and can capture and/or produce output data in the form of evidence.

Both of these concepts, attestations and evidence can be incorporated into one or more of the layers, resulting in a richer, more communicative schema and one that affords the functionality that is common in compliance assessment and audit use cases.

# Next Steps

- [ ] Propose schema changes and accompanying documentation via PR
- [ ] Socialize proposed schema changes with impacted community members
- [ ] Generate new Go types to represent the evidence and attestation concepts

eddie-knight:
> "attestations and evidence can be incorporated into one or more of the layers"

Hey @trumant, could you elaborate on this? The evidence gathered and attestation results are both useful in the audit activities, and potentially during enforcement, but I don't anticipate any changes would be necessary for that. 

Privateer adds "payload" as an interface / un-typed object... perhaps we could add that in the layer 4 schema?

trumant:
@eddie-knight See https://github.com/trumant/pvtr-github-repo/blob/redesign-payload-interface/architecture.md#input-evidence-data-from-multiple-sources-may-duplicate-andor-make-conflicting-assertions and the sections that follow that speak to this.

My claim, broadly, is that layer 4 should evolve, specifically that [AssessmentRequirement](https://github.com/revanite-io/sci/blob/main/schemas/layer-2.cue#L78) should expand to have a notion of AssessmentMethods that capture input data to the assessment process and also capture the evidence/facts that the method used to reach a conclusion.

My belief is that exporting this richer, more detailed data from Layer 4 will dramatically improve the capabilities of Layers 5 and 6.

xee5ch:
> Layer 4 assessments evaluate input data which could include attestations of compliance to one or more controls and their requirements and can capture and/or produce output data in the form of evidence.

Maybe to step back a bit, what do we specifically mean by attestation of compliance to one or more controls? Some standards and developer communities consider attestation and endorsement separate things, but it is hard to know what is meant by attestation here.

I am very interested in the design piece and will check out @trumant's attestation architecture spec in the branch. Is the place to provide relevant feedback about it?

eddie-knight:
Sure, I'm curious to get more thoughts here

trumant:
> what do we specifically mean by attestation of compliance to one or more controls

[...]

PR #83, opened 2025-08-08
feat!: updates AssessmentStep to AssessmentMethod

### LINKED ISSUE #73: Support mapping declarative policy identifiers or locations to `layer4.Assessments` as steps
# Overview

Adding this linkage is would support use cases where declarative policy serves as the external source of truth for the desired state, while the function can provide the execution logic for how to verify that state against input from live environments. Users could define the assessment declaratively in a `pre-run` state via Layer4 YAML.

# Next Steps

- [x] Add a wrapper struct around `AssessmentSteps`
- [x] Add an optional AssessmentSteps `id` field

> Note: Overlaps with proposed `AssessmentMethod`  issues #23, but only covers a subset of that issue scope

jpower432:
# What Changed? 

- `AssessmentMethod` was added
- The `cue` schema for `layer4` was updated to validate the updated structure generate using the library

:warning: **BREAKING**
This removes the `AssessmentStep` type. Similar functionality can be acheived with the `MethodExecutor` on `AssessmentMethod`. 

> Note: Adapted from work on `AssessmentMethods` done by @trumant in #19 

## Rationale

Adding AssessmentMethod supports a multi-method Assessment that is flexible enough to accommodate different types of procedures to complete a control requirement. This new structure also captures detailed information about each individual check. The goal of `AssessmentMethod` is to be self-contained, allowing for composability and ensuring the solution retains the core functionality and structure of AssessmentSteps.

- Is Assessment Method the most clear terminology to use? Perhaps Assessment Procedure?
- If we want Assessment Methods to be seen as only as "options" do they need individual pre-condition steps? 

## Testing

A new file under `layer4/testdata` was added that is populated output from a Layer 4 structure generated by the `TestEvaluate` unit test. This was used to validate against the updated in the `cue` schema.

`cue vet -d "#Layer4" schemas/layer-4.cue layer4/testdata/good-evaluation.yml`

## Issues
Closes #73
Overlaps/Impact #23


schemas/layer-4.cue
```diff
@@ -1,37 +1,93 @@
...
+	"control-id": string
+	// Result is the overall result of the control evaluation
+	result: #Result
+	// Message is the human-readable result of the final assessment to run in this evaluation
+	message: string
+	// CorruptedState is true if the control evaluation was interrupted and changes were not reverted
+	"corrupted-state": bool
+	// RemediationGuide is the URL to the documentation for this evaluation
+	"remediation-guide": string
+	// Assessments is a map of pointers to Assessment objects to establish idempotency
```
trumant: This looks to be a slice not a map

schemas/layer-4.cue
```diff
@@ -1,37 +1,93 @@
...
+	// Description is a human-readable description of the test
+	description: string
+	// Result is true if the test passed
+	result: #Result
+	// Message is the human-readable result of the test
+	message: string
+	// Methods is a slice of assessment methods that were executed during the test
+	methods: [...#AssessmentMethod]
+	// MethodsExecuted is the number of assessment methods that were executed during the test
+	"methods-executed"?: int @go(MethodExecuted)
```
trumant: Singular/plural mismatch

schemas/layer-4.cue
```diff
@@ -1,37 +1,93 @@
...
 #Change: {
-    "target-name": string
-    applied: bool
-    reverted: bool
-    error?: string
-    "target-object"?: _
-}
\ No newline at end of file
+	// TargetName is the name or ID of the resource or configuration that is to be changed
+	"target-name": string @go(TagertName)
```
trumant:
```suggestion
	"target-name": string @go(TargetName)
```

trumant (commented):
Quick review from my phone. Will take a deeper look on Monday morning

schemas/layer-4.cue
jpower432: Good catch! Found that in a couple places actually. Updated on the last commit.

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
 package schemas
 
 #Layer4: {
-    evaluations: [#ControlEvaluation, ...#ControlEvaluation]
+	evaluations: [#ControlEvaluation, ...#ControlEvaluation]
 }
 
-// Types
-
+// ControlEvaluation is a struct that contains all assessment results, organized by name.
```
trumant:
```suggestion
// ControlEvaluation provides all assessment results for a single control
```

https://tip.golang.org/doc/comment#type is helpful guidance, but also my take is the cue schema and its comments should be consumable by itself, without needing to understand golang types or even be aware that there are Golang "bindings" associated with the schema.

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
...
-    name: string
-    "control-id": string
-    result: #Result
-    message: string
-    "documentation-url"?: =~"^https?://[^\\s]+$"
-    "corrupted-state"?: bool
-    "assessment-results"?: [...#AssessmentResult]
+	// Name is the name of the control being evaluated
+	name: string
+	// ControlID is the unique identifier for the control being evaluated
```
trumant:
```suggestion
	// ControlID uniquely identifies the control
```

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
...
+	"control-id": string
+	// Result is the overall result of the control evaluation
```
trumant:
```suggestion
	// Result communicates whether the evaluation has been run, and if so, the outcome(s)
```

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
...
+	result: #Result
+	// Message is the human-readable result of the final assessment to run in this evaluation
```
trumant:
```suggestion
	// Message describes the result of the evaluation
```

Maybe...

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
...
+	message: string
+	// CorruptedState is true if the control evaluation was interrupted and changes were not reverted
+	"corrupted-state"?: bool
```
trumant, line 18: This is a characteristic of the system(s) being evaluated in the assessment. I'd like to cordon off this implementation/technical detail of the system under assessment and the assessment technical implementation so it doesn't detract and confuse a more compliance-oriented audience who may not have any interest or understanding of these technical details and simply want the "did the control pass or not" visibility.

Perhaps we model new types that can encapsulate these 2 concerns of: system under eval properties/issues and system doing the eval properties/issues.

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
...
+	// RemediationGuide is the URL to the documentation for this evaluation
```
trumant:
```suggestion
	// RemediationGuide provides a URL with guidance on how to remediate systems that fail control evaluation
```

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
...
+	"remediation-guide"?: string
+	// Assessments is a slice of pointers to Assessment objects to establish idempotency
```
trumant:
```suggestion
	// Assessments represents the collection of results from evaluation of each control requirement
```

schemas/layer-4.cue
jpower432: Thanks for the feedback @trumant. I think you raise a really good point. There are quite a few places where the Go implementation is referenced in the comments. I can certainly go back and refine the comments with this additional feedback, but I'm also wondering if we should have the comments on the Go types diverge from the comments in the schema. Do you think the Go type doc comments be altered as well?

I think your feedback highlights a related point: it would be helpful to document how Layer 4 is (should be) modified for contributors (i.e. `modify schema` -> `generate types`? -> `manually copy or refactor layer 4 changes`?). This seems like a great topic for a follow-on issue.

schemas/layer-4.cue
jpower432: This is an interesting point @trumant. I am wondering what a new set of types might look like. My interpretation of this field and the associated `Change` is that the assessment tool applies pre-configured changes to system under assessment to trigger a specific outcome (like a test fixture). If a failure occurs in the teardown of this fixture, the `corrupted-state` field would be set to true, and this would then be reflected in a 'Needs Review' result and an error message.

Since this was a pre-existing field in the schema, would it be worth exploring in a new issue?

trumant: Agreed, that these changes can be explored elsewhere

schemas/layer-4.cue
```diff
@@ -1,37 +1,101 @@
...
+	// Description provides a detailed explanation of the assessment
+	description: string
+	// Result communicates whether the assessment has been run, and if so, the outcome(s)
+	result: #Result
+	// Message describes the result of the assessment
+	message: string
+	// Methods defines the assessment methods associated with the assessment
+	methods: [...#AssessmentMethod]
+	// MethodsExecuted is the number of assessment methods that were executed during the assessment
+	"methods-executed"?: int @go(MethodsExecuted)
```
trumant: Why is this data summarized here when it can be computed easily enough by the data consumer?

jpower432: Thanks for the feedback. You're right - the `steps-counter` field made sense for a standalone function, but with the new `run` field in the AssessmentMethod struct, it's redundant.

jpower432:
Add a commment to describe the refactor in https://github.com/ossf/gemara/pull/83/commits/d431b80c5a8d0469f53a72e820632b9b28cc5e9c
# What Changed? 

- `AssessmentProcedure` was added
- `AssessmentStep` functionality is kept as is and moved under the `Assessment Procedure`

## Rationale

Results from a maintainer discussion where we decided the assessment logic should keep steps as they currently are, but adding procedure to allow for each assessment to contain multiple groups of steps.

## Points of Discussion

- The concept of `ProcedureMethod` was added here to ensure the automated control evaluation skipped over any assessment procedures that should be completed by a human. I initially added the `Test` method and the `Observation` method to describe how the results of the procedure are to be determined.

- The Assessment does not halt other procedures if a procedure fails. However, the procedure will halt if a step fails.

## Open Questions

- Message handling - If mulitple procedures run, how should should the `Assessment` message aggregate them?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
+// ControlEvaluation provides all assessment results for a single control
 #ControlEvaluation: {
-    name: string
-    "control-id": string
-    result: #Result
-    message: string
-    "documentation-url"?: =~"^https?://[^\\s]+$"
-    "corrupted-state"?: bool
-    "assessment-results"?: [...#AssessmentResult]
+	// Name is the name of the control being evaluated
```
trumant, line 9: If we expect a tool will consume lower layers and produce output conforming with `ControlEvaluation`, should we consider a datetime here or elsewhere?

jpower432: I agree a timestamp would be valuable. The `run-duration` field in Assessments is useful, but I'm wondering if it would make more sense to include a `start` and `end` datetime and let the consumer calculate the duration as needed. WDYT?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
+	// Description provides a detailed explanation of the assessment
+	description: string
+	// Result communicates whether the assessment has been run, and if so, the outcome(s)
+	result: #Result
+	// Message describes the result of the assessment
+	message: string
+	// Procedures defines the assessment procedures associated with the assessment
+	procedures: [...#AssessmentProcedure]
+	// RunDuration is the time it took to run the assessment
+	"run-duration"?: string @go(RunDuration)
```
trumant, line 40: Why do the typical data consumers care how long the assessment took? Or, put another way, what value does this data provide and to whome?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
+	// Value is the object that was returned during the assessment
+	value?: _
```
trumant, line 42: What is this?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
+// Assessment provides all assessment results from evaluation of a single control requirement
+#Assessment: {
+	// RequirementID uniquely identifies the requirement being tested
+	"requirement-id": string @go(RequirementId)
+	// Applicability provides identifier strings to determine when this assessment is applicable
+	applicability: [...string]
+	// Description provides a detailed explanation of the assessment
+	description: string
+	// Result communicates whether the assessment has been run, and if so, the outcome(s)
+	result: #Result
```
trumant, line 34: The "has been run" language here confuses me.

In what circumstances or use cases do we expect to have a document conforming to this schema that does not represent an Assessment activity that occurred in the past?

Is there value in having a layer4 payload that communicates an "intended" or "future" assessment?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
+	// Message describes the result of the assessment
+	message: string
```
trumant, line 36: If it's describing the result, can we put it in `#Result`?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
 
-#Result: "Passed" | "Failed" | "Needs Review"
+// Result describes valid assessment outcomes before and after execution.
+#Result: #ResultWhenRun | "Not Run"
 
+// Result describes the outcome(s) of an assessment procedure when it is executed.
+#ResultWhenRun: "Passed" | "Failed" | "Needs Review" | "Not Applicable" | "Unknown"
+
+// ProcedureMethod describes method options that can be used to determine the results
+#ProcedureMethod: "Test" | "Observation"
```
trumant, line 87: I think we need to clarify the language here. In my experience testing involves observation.

What value does this provide to the intended data consumers? How do you see them using this fact?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
+	// Method describe the high-level method used to determine the results of the procedure
+	method: #ProcedureMethod
+	// Run is a boolean indicating whether the procedure was run or not. When run is true, result is expected to be present
+	run: bool
+	// RemediationGuide provides a URL with remediation guidance associated with the control's assessment requirement and this specific assessment procedure
+	"remediation-guide"?: #URL @go(RemediationGuide)
+	// Documentation provides a URL to documentation that describes how the assessment procedure evaluates the control requirement
+	documentation?: #URL
+	// Steps provides the address for the assessment steps executed
+	"steps"?: [...string]
```
trumant, line 64: How is a consumer of this data going to interpret, analyze or otherwise make use of this particular field?

schemas/layer-4.cue
```diff
@@ -1,37 +1,108 @@
...
+	message!: string
+	result!: #ResultWhenRun
 }
 
-#Result: "Passed" | "Failed" | "Needs Review"
+// Result describes valid assessment outcomes before and after execution.
+#Result: #ResultWhenRun | "Not Run"
 
+// Result describes the outcome(s) of an assessment procedure when it is executed.
+#ResultWhenRun: "Passed" | "Failed" | "Needs Review" | "Not Applicable" | "Unknown"
```
trumant, line 84: Contingent on the decision above on "has been run", but I'd propose removing `#ResultWhenRun` altogether

schemas/layer-4.cue
eddie-knight: this one's a bit legacy — was used to log details about the inspected value

schemas/layer-4.cue
jpower432: Agree. This was part of the attempt to align the schema with the current struct in the Go library. I would propose start and end timestamps instead.

schemas/layer-4.cue
jpower432: I think there are two use cases I was trying to capture:

1. Currently, steps can have the state of `NotRun` if a prior Assessment Step has failed and the process is halted.
2. Part of my intent/goal with adding the `Procedures` was to allow Layer4 to be used in pre-run state (essentially like a test plan) as an input to a tool to execute the tests and fill in the result information.

schemas/layer-4.cue
jpower432: I agree, these could be more descriptive. I wanted to allow categorization of the prodecures based on if the results were determined with automation or human judgement of behavior or state.

Perhaps:
```suggestion
#EvaluationMethod: "Automated" | "Manual"
```

How this would be used might depend on if we want to support Layer4 as an input. This allows categorization of the procedures so that they might be handled differently to determine the results.

jpower432:
Looks like this branch has drifted from main. I will perform a rebase and get it back into a reviewable state.

schemas/layer-4.cue
jpower432: I'm thinking a potential solution here is to capture whether a step, assessment, or procedure ran in different field like `run` and keep the results field focused on assessment outcomes. Perhaps this would align with your comment about adding `message` to results. WDYT?

schemas/layer-4.cue
jpower432: I proposed something to address the `run-duration` field here - https://github.com/ossf/gemara/pull/112

schemas/layer-4.cue
jpower432: I proposed something here to update the existing `Result` field - https://github.com/ossf/gemara/pull/113

jpower432:
Superseded by #117

PR #88, opened 2025-08-13
feat: Added ControlFamily.Id

### LINKED ISSUE #87: [PROPOSAL] Create an ID Field in the Control Family Struct
## Problem

I'm working on a PR to move the oscal parser from security-baseline into gemara layer 2. When I saw how the [OSCAL group ID was being populated](https://github.com/ossf/security-baseline/blob/054f61119a8cc9fba48c7348fae65ea3583b925b/cmd/pkg/baseline/generator_oscal.go#L53), I noticed that it's fabricating a new term that's not part of the gemara catalog. 

## Potential Solution

Create an ID Field in the Control Family struct.

eddie-knight:
Resolves #87

### DIFF
--- a/layer2/generated_types.go
+++ b/layer2/generated_types.go
@@ -56,6 +56,8 @@ type MappingReference struct {
 }
 
 type ControlFamily struct {
+	Id	string	`json:"id" yaml:"id"`
+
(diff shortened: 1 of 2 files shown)


trumant (approved):
Dig it, good catch and fix

Issue #86, opened 2025-08-12
Increase fidelity of mapping relationships


eddie-knight:
It has been suggested that we prepare future to support conversion with the OSCAL mapping model by adding a value to mapping relationships which captures the strength of the relationship on a sliding scale (ie. 1 to 10)

## Current

Mapping relationships currently look like this in practice:

```yaml
guideline-mappings:
  - reference-id: NIST-CSF
    identifiers:
      - PR.DS-1
  - reference-id: CCM
    identifiers:
      - DCS-04
      - DCS-06
  - reference-id: ISO_27001
    identifiers:
      - 2013 A.10.1.1
  - reference-id: NIST_800_53
    identifiers:
      - SC-28
      - SC-29
      - SC-30
```

## Proposed Solution

This approach would add fidelity at the expense of brevity:

```yaml
guideline-mappings:
  - reference-id: NIST-CSF
    entries:
      - entry-id: PR.DS-1
        strength: 3
  - reference-id: CCM
    entries:
      - entry-id: DCS-04
        strength: 4
      - entry-id: DCS-06
        strength: 2
  - reference-id: ISO_27001
    entries:
      - entry-id: 2013 A.10.1.1
        strength: 6
  - reference-id: NIST_800_53
    entries:
      - entry-id: SC-28
        strength: 5
      - entry-id: SC-29
        strength: 2
      - entry-id: SC-30
        strength: 1
```

jpower432:
Adding a couple thoughts:

- Is the goal to provide this field for just Layer 1 to Layer 1 `guideline-mappings` or Layer 2 to Layer 1 `guideline-mappings` as well? I think Layer 2 to Layer 1 could have different considerations.

- I think adding an optional `remarks` or a similar type of field would be useful so an author can add context on the their reasoning on the strength value. Thoughts?

eddie-knight:
Sure — if we're going to the trouble of making it a struct, another optional field won't grind my gears.

Right now I'm thinking about Layer 2, made the potentially naive assumption that L1 would be pretty much the same.

jpower432:
Ack. Starting with Layer 2 makes sense. Seems `strength` could be a useful metric for both. I was thinking of L1-to-L1 mappings as crosswalks demonstrating equivalence and L2-to-L1 mappings demonstrating fulfillment or coverage of a guideline. So perhaps `strength` might be measuring different things dependening on the layer. Does that align with how you're thinking about it?

eddie-knight:
[...]

PR #90, opened 2025-08-14
feat: Mapping identifiers are now entry objects

eddie-knight:
This PR relates to #86, but does not address the potential for a rename of Layer 3 mappings.

Another thing that is out of scope for this PR... It occurs to me that we could be adding comments to CUE that will get auto-generated into the go docs (and shown in the IDE for dev-users)

### DIFF
--- a/layer2/test-data/good-ccc.yaml
+++ b/layer2/test-data/good-ccc.yaml
@@ -36,23 +36,29 @@ control-families:
           integrity and confidentiality.
         threat-mappings:
           - reference-id: CCC
-            identifiers:
-              - CCC.TH02 # Data is Intercepted in Transit
+            entries:
+              - reference-id: CCC.TH02
+                strength: strong
+                remarks: Data is Intercepted in Transit
         guideline-mappings:
           - reference-id: CSF
-            identifiers:
-              - PR.DS-02 # Data-in-transit is protected
+            entries:
+              - reference-id: PR.DS-02
+                strength: strong
+                remarks: Data-in-transit is protected
           - reference-id: CCM
-            identifiers:
-              - IVS-03
-              - IVS-07
+            entries:
+              - reference-id: IVS-03
+              - reference-id: IVS-07
           - reference-id: ISO-27001
-            identifiers:
-              - 2013 A.13.1.1
+            entries:
+              - reference-id: 2013 A.13.1.1
+                strength: strong
+                remarks: This control is closely related to 2013 A.13.1.1.
           - reference-id: NIST-800-53
-            identifiers:
-              - SC-8
-              - SC-13
+            entries:
+              - reference-id: SC-8
+              - reference-id: SC-13
         assessment-requirements:
           - id: CCC.C01.TR01
             text: |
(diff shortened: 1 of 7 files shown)


schemas/layer-2.cue
```diff
@@ -70,9 +70,17 @@ package schemas
...
+
 #Mapping: {
-    "reference-id": string @go(ReferenceId)
-    identifiers: [...string]
+	"reference-id": string @go(ReferenceId)
+	entries: [...#MappingEntry]
+	remarks?: string
+}
+
+#MappingEntry: {
```
trumant, line 80: I feel like we want to avoid duplicating types across the schema and go packages.

Perhaps there is a way to define each layer schema as a collection of cue files where one or more of those files may contain types like those one which is reused across multiple layers

schemas/layer-1.cue
```diff
@@ -118,11 +118,17 @@ package schemas
...
 	"reference-id": string @go(ReferenceId)
-	identifiers: [...string]
+	entries: [...#MappingEntry]
 	// Adding context about this particular mapping and why it was mapped.
 	remarks?: string
 }
 
+#MappingEntry: {
+	"reference-id": string @go(ReferenceId)
+	strength: string
```
trumant: What is this? Can we enumerate its values in a closed set?

schemas/layer-2.cue
eddie-knight: I agree, and started looking into this before realizing I was getting too distracted from the point at hand.

Opened an issue for this: https://github.com/revanite-io/gemara/issues/91

schemas/layer-1.cue
eddie-knight: Strength is the focus of the issue this resolves, #86. It was erroneously a string instead of int, but that's been corrected in a recent commit.

schemas/layer-1.cue
```diff
@@ -118,11 +118,17 @@ package schemas
...
 	"reference-id": string @go(ReferenceId)
-	identifiers: [...string]
+	entries: [...#MappingEntry]
 	// Adding context about this particular mapping and why it was mapped.
 	remarks?: string
 }
 
+#MappingEntry: {
+	"reference-id": string @go(ReferenceId)
+	strength: int
```
jpower432: Wondering we should define a boundary here for clarity. It would be easier to document what the numbers or numbers ranges mean if we do this.
```suggestion
	strength: int & >=1 & <=10
```

layer2/test-data/good-ccc.yaml
```diff
@@ -36,23 +36,29 @@ control-families:
...
-              - PR.DS-02 # Data-in-transit is protected
+            entries:
+              - reference-id: PR.DS-02
+                strength: 7
+                remarks: Data-in-transit is protected
           - reference-id: CCM
-            identifiers:
-              - IVS-03
-              - IVS-07
+            entries:
```
jpower432, line 52: Some of these entries are failing when you vet them against the schema, but it is for multiple reasons (i.e. missing control family `id`) . It may be worth just opening a new issue to avoid blocking this.

eddie-knight: ughhhhh, good catch. I'm gonna strip down the size of the test-data yaml. 

It was originally compiled from a compatible version of a CCC catalog, but the file is so large that even Copilot and Cursor Pro are failing to update it due to the size.

schemas/layer-2.cue
```diff
@@ -70,9 +70,17 @@ package schemas
...
+	"reference-id": string @go(ReferenceId)
+	strength: int
```
jpower432: To match layer 1
```suggestion
	strength: int & >=1 & <=10
```

eddie-knight: I'm not managing to get this structure passing with cue vet yet 😬 

```sh
$ cue vet -d '#Catalog' ./layer2/test-data/good-ccc.yaml schemas/layer-2.cue
"control-families".0.controls.0."guideline-mappings".1.entries.0.strength: incomplete value >=1 & <=10 & int
"control-families".0.controls.0."guideline-mappings".1.entries.1.strength: incomplete value >=1 & <=10 & int
"control-families".0.controls.0."guideline-mappings".3.entries.0.strength: incomplete value >=1 & <=10 & int
"control-families".0.controls.0."guideline-mappings".3.entries.1.strength: incomplete value >=1 & <=10 & int
"control-families".0.id: incomplete value string:
    ./schemas/layer-2.cue:35:9
metadata.title: incomplete value string:
    ./schemas/layer-2.cue:20:12
```

jpower432: Just tested locally with your last commit. Passing now!

jpower432 (approved):
LGTM

trumant (approved):

PR #92, opened 2025-08-14
feat: add OSCAL export functionality to layer2

zohayb23:
<details>
<summary>Adds OSCAL export functionality to layer2 package</summary>

This PR implements OSCAL (Open Security Controls Assessment Language) export functionality, allowing Catalog objects to be converted to OSCAL format.

</details>

## Summary
Adds OSCAL export functionality to the layer2 package, allowing Catalog objects to be converted to OSCAL format.

## Changes
- `ToOSCAL()` method on `Catalog` struct for OSCAL conversion
- Added `github.com/defenseunicorns/go-oscal` dependency
- Comprehensive test coverage

## Example

I used this logic to convert the following Gemara content into OSCAL:

<details>
<summary>Gemara (input)</summary>

```yaml
metadata:
  title: "Test Security Catalog"
controlFamilies:
  - id: "AC"
    title: "Access Control"
    controls:
      - id: "AC-01"
        title: "Access Control Policy"
```

</details>

<details>
<summary>OSCAL (output)</summary>

```json
{
  "catalog": {
    "metadata": {
      "title": "Test Security Catalog",
      "version": "devel"
    },
    "groups": [
      {
        "id": "AC",
        "title": "Access Control",
        "controls": [
          {
            "id": "AC-01",
            "title": "Access Control Policy"
          }
        ]
      }
    ]
  }
}
```

</details>

## Files Changed
- `layer2/oscal-parser.go` - OSCAL conversion logic
- `layer2/oscal-parser_test.go` - Test coverage
- `go.mod` - Added dependency
- `go.sum` - Updated checksums


layer2/oscal-parser.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	oscal "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-3"
+)
+
+// ToOSCAL converts a Catalog to OSCAL Catalog format.
+func (c *Catalog) ToOSCAL(controlFamilyIDs map[string]string,
+	versionOSPS, controlHREF, catalogUUID, namespace string) (oscal.Catalog, error) {
```
jpower432: Just a nit here, assuming this would support Layer 2 catalog exports broadly.
```suggestion
	version, controlHREF, catalogUUID, namespace string) (oscal.Catalog, error) {
```

layer2/oscal-parser.go
```diff
@@ -0,0 +1,93 @@
...
+
+	now := time.Now()
+	oscalCatalog := oscal.Catalog{
+		UUID:   catalogUUID,
```
jpower432: I think passing in the UUID instead of generating it makes this more testable, but I am wondering how it would be handled if the method starts covering other OSCAL elements that require a UUID. You may have already come accross this, but `go-oscal` has a couple functions that can help with random and deterministic [UUID generation](https://github.com/defenseunicorns/go-oscal/tree/main/src/pkg/uuid).

layer2/oscal-parser.go
```diff
@@ -0,0 +1,93 @@
...
+		Groups: nil,
+		Metadata: oscal.Metadata{
+			LastModified: now,
```
jpower432: The Layer 2 catalog also has a `LastModified` string in the `Metadata` that could be parsed and used here. Perhaps using the current time could be a fallback if that is not set?

layer2/oscal-parser.go
```diff
@@ -0,0 +1,93 @@
+package layer2
```
trumant: The file naming here `oscal-parser.go` makes me assume that this contains code that consumes OSCAL. However, the code here is more about producing OSCAL. Perhaps rename to `oscal_generator.go`

Also, note I specified an underscore, not a hyphen as all other `.go` files in these directories use underscores in their names, rather than hyphens.

layer2/oscal-parser_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+			if (err == nil) == tt.wantErr {
+				t.Errorf("ToOSCAL() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			if tt.wantErr {
+				return
+			}
+
```
trumant: It feels like we should perhaps be validating that the OSCAL produced is valid according to https://github.com/defenseunicorns/go-oscal/blob/main/src/pkg/validation/validator.go#L137

WDTY?

layer2/oscal-parser.go
```diff
@@ -0,0 +1,93 @@
...
+			Title:        c.Metadata.Title,
+			Version:      versionOSPS,
+		},
+	}
+
+	catalogGroups := []oscal.Group{}
+
+	for _, family := range c.ControlFamilies {
+		group := oscal.Group{
+			Class:    "OSPS",
```
jpower432: I would suggest `family` for `Class`. This aligns with how the 800-53 catalog defines control families.

jpower432 (commented):
Adding some initial comments here. I agree with @trumant that validation with the schema would be helpful. Especially to catch things like invalid control strings.

layer2/oscal-parser_test.go
zohayb23: @trumant that makes sense; I will add validation according to validator.go

layer2/oscal-parser.go
zohayb23: Thanks for the suggestion @jpower432. I've added a TODO comment to consider using uuid.NewUUID() and uuid.NewUUIDWithSource() when we start generating UUIDs for individual controls and assessment requirements.

layer2/oscal-parser.go
zohayb23: You're right; I've implemented the fallback logic to use the catalog's LastModified string when available, with the current time as fallback if it's not set or invalid.

eddie-knight:
@zohayb23 can you please sign your commits so that the DCO check passes?

The two options are to do it via rebase, or if you're content squashing the git history, you can redo your commits by resetting `git reset upstream/main` then `git commit -sm "<your message>"` (the `-s` adds the basic signature)

eddie-knight:
@jpower432 @trumant — Did all of yalls feedback get merged in? If so, I'll stamp and merge this once Zobi gets the DCO check passing

go.mod
```diff
@@ -5,12 +5,15 @@ go 1.23.0
 toolchain go1.24.1
 
 require (
+	github.com/defenseunicorns/go-oscal v0.6.2
```
trumant, line 8:
```suggestion
	github.com/defenseunicorns/go-oscal v0.6.3
```

See https://github.com/defenseunicorns/go-oscal/releases/tag/v0.6.3

layer2/oscal_generator.go
```diff
@@ -0,0 +1,101 @@
...
+	"time"
+
+	oscal "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-3"
+)
+
+// ToOSCAL converts a Catalog to OSCAL Catalog format.
+// TODO: Consider using go-oscal's UUID generation for future OSCAL elements:
+// - uuid.NewUUID() for random UUIDs in production
+// - uuid.NewUUIDWithSource() for deterministic UUIDs in testing
+func (c *Catalog) ToOSCAL(controlFamilyIDs map[string]string,
```
trumant, line 15: I can't figure out what these function parameters do when reading the function comment. Please add details to the comment that explain how a caller should use those parameters to achieve the desired function return value

layer2/oscal_generator.go
```diff
@@ -0,0 +1,101 @@
...
+		Groups: nil,
+		Metadata: oscal.Metadata{
+			LastModified: lastModified,
+			Links: &[]oscal.Link{
+				{
+					Href: fmt.Sprintf(controlHREF, version, ""),
+					Rel:  "canonical",
+				},
+			},
+			OscalVersion: "1.1.3",
```
trumant, line 35: Does the oscal module not define this version as a constant already that we could reference here?

jpower432:
@eddie-knight All of my feedback is addressed. There might need to be some alignment between this work and the draft work in #99, but I think that can be addressed iteratively post-merge.

eddie-knight (approved):
Excellent! Thanks for applying the feedback and DCO changes @zohayb23 🚀

zohayb23:
@eddie-knight sure thing! I appreciate @trumant and @jpower432 for the feedback as well.

layer2/oscal_generator.go
jpower432: It does not, but looks like there is an open issue - https://github.com/defenseunicorns/go-oscal/issues/146

PR #98, opened 2025-08-15
chore(deps): Bump ospo-reusable-workflows/auto-labeler to 23b65d1

eddie-knight:


trumant:
What problem do we expect this to address?

eddie-knight:
In a dream world, this will fix the fact that tagging is only happening for `maintenance` PRs, and it'll magically start working for everything else again.

Realistically, this just rules that out as a possibility. Update to latest before complaining about bugs, etc.

trumant (approved):

PR #103, opened 2025-08-20
chore: improve Makefile and CI workflow

trumant:
This PR simplifies the Makefile significantly to remove unneeded targets. The CI workflow is simplified to remove the install of `golint` as `golangci-lint` is already run in the lint workflow


jpower432 (approved):
LGTM

PR #106, opened 2025-08-21
fix: Simplify ToOSCAL function to use internal IDs/version, generate UUID internally

zohayb23:
## Changes
- Use internal ControlFamily.Id instead of external mapping
- Use internal c.Metadata.Version instead of parameter
- Generate UUID internally instead of requiring it
- Remove incorrect namespace documentation
- Update tests to match the new signature

### DIFF
--- a/layer2/oscal_generator.go
+++ b/layer2/oscal_generator.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


eddie-knight:
Renaming the title now as a bit of a nitpick... Refactoring is a non-functional change, only modifying internal logic without change to inputs or outputs.

eddie-knight (approved):

PR #107, opened 2025-08-21
fix: more revanite-io to ossf renames

trumant:
This PR updates remaining references to `revanite-io/gemara` and `revanite-io/sci` to be `ossf/gemara`


jpower432 (approved):
LGTM

PR #109, opened 2025-08-21
fix: Move control remediation guide to assessment recommendation

eddie-knight:
Labelling this as a fix rather than a feat because we don't have any documented use of this field in its current location, but we have an end-user request for assessment requirement recommendations to be embedded alongside assessment results.

This also fixes the cue schema to bring it up to date with the go package.


schemas/layer-4.cue
```diff
@@ -1,37 +1,43 @@
...
-	applied:          bool
-	reverted:         bool
+	"target-name":    string @go(TargetName)
+	description:      string
+	"target-object"?: _ @go(TargetObject)
+	applied?:         bool
+	reverted?:        bool
 	error?:           string
-	"target-object"?: _
+	allowed?:         bool
```
eddie-knight:
```suggestion
```

schemas/layer-4.cue
```diff
@@ -1,37 +1,42 @@
 package schemas
 
-#Layer4: {
-	evaluations: [#ControlEvaluation, ...#ControlEvaluation]
+#EvaluationOutput: {
+	"evaluation-set": [#ControlEvaluation, ...#ControlEvaluation] @go(EvaluationSet)
+	...
 }
```
eddie-knight: IIRC, I think this aligns with the recent discussion around the possibility of separating the evaluation plan and results.

I went with "output" arbitrarily, but looking back I could see a case for naming this `EvaluationResult`

```suggestion
#EvaluationResults: {
	"evaluation-set": [#ControlEvaluation, ...#ControlEvaluation] @go(EvaluationSet)
	...
}
```

eddie-knight: @jpower432 curious which one you prefer here.

jpower432: I would lean toward "results" to keep the term precise.

eddie-knight: Applied!

layer4/assessment.go
```diff
@@ -31,6 +31,8 @@ type Assessment struct {
 	Value interface{} `yaml:"value,omitempty"`
 	// Changes is a slice of changes that were made during the test
 	Changes map[string]*Change `yaml:"changes,omitempty"`
+	// Recommendation should optionally present the corresponding text from a layer 2 assessment requirement
+	Recommendation string `yaml:"recommendation,omitempty"`
```
jpower432, line 35: I like this change because I think the closer the remediation data is to the specific assessment logic, the more actionable and precise it can be. My comment is more a nit/question about the field comment - do you envision this field being dedicated to duplicating the Layer 2 text or can it be more flexible than that?

eddie-knight: Let me pop over to that line and suggest a revision

layer4/assessment.go
```diff
@@ -31,6 +31,8 @@ type Assessment struct {
 	Value interface{} `yaml:"value,omitempty"`
 	// Changes is a slice of changes that were made during the test
 	Changes map[string]*Change `yaml:"changes,omitempty"`
+	// Recommendation should optionally present the corresponding text from a layer 2 assessment requirement
```
eddie-knight:
```suggestion
	// Recommendation is a string to aid users in remediation, such as the text from a layer 2 assessment requirement
```

cc/ @jpower432

jpower432 (approved):

PR #110, opened 2025-08-22
chore: generate layer4 types from cue schema

trumant:
This change updates the project's Makefile to use
`cue gengotypes` to generate the layer4 struct types


eddie-knight:
I think this will be possible after we resolve the `changes` discussion

trumant:
> I think this will be possible after we resolve the `changes` discussion

Yep, I'm raising it to hurry that discussion along.

PR #113, opened 2025-08-22
feat!: groups assessment outcomes into Results (for discussion)

jpower432:
## Objective
Adds schema and test data updates to propose an update to how `results` are described in Layer 4.

> Added initially for discussion and if the proposed solution is accepted, I can implement in the library.

## What Changed?

- A `run` field is introduced to capture if the Evaluation or Assessment is in a `pre-run` or `post-run` state.
- Groups results related fields into a `Results` object. All known result states except `Not Run` are added to `Status` and message is added.

Related discussion: https://github.com/ossf/gemara/pull/83#discussion_r2293620583

### DIFF
--- a/layer4/test-data/pvtr-baseline-scan.yaml
+++ b/layer4/test-data/pvtr-baseline-scan.yaml
@@ -4,8 +4,10 @@ payload: omitted
 evaluation-set:
   - name: ""
     control-id: OSPS-AC-01
-    result: Passed
-    message: Two-factor authentication is configured as required by the parent organization
+    run: true
+    result:
+      state: Passed
+      message: Two-factor authentication is configured as required by the parent organization
     corrupted-state: false
     assessments:
     - requirement-id: OSPS-AC-01.01
@@ -14,8 +16,10 @@ evaluation-set:
       - Maturity Level 2
       - Maturity Level 3
       description: When a user attempts to access a sensitive resource in the project's version control system, the system MUST require the user to complete a multi-factor authentication process.
-      result: Passed
-      message: Two-factor authentication is configured as required by the parent organization
+      run: true
+      result:
+        state: Passed
+        message: Two-factor authentication is configured as required by the parent organization
       steps:
       - github.com/revanite-io/pvtr-github-repo/evaluation_plans/osps/access_control.orgRequiresMFA
       steps-executed: 1
@@ -24,8 +28,10 @@ evaluation-set:
       changes: {}
   - name: ""
     control-id: OSPS-AC-02
-    result: Passed
-    message: This control is enforced by GitHub for all projects
+    run: true
+    result:
+      state: Passed
+      message: This control is enforced by GitHub for all projects
     corrupted-state: false
     assessments:
     - requirement-id: OSPS-AC-02.01
@@ -34,8 +40,10 @@ evaluation-set:
       - Maturity Level 2
       - Maturity Level 3
       description: When a new collaborator is added, the version control system MUST require manual permission assignment, or restrict the collaborator permissions to the lowest available privileges by default.
-      result: Passed
-      message: This control is enforced by GitHub for all projects
+      run: true
+      result:
+        state: Passed
+        message: This control is enforced by GitHub for all projects
       steps:
       - github.com/revanite-io/pvtr-github-repo/evaluation_plans/reusable_steps.GithubBuiltIn
       steps-executed: 1
@@ -44,8 +52,10 @@ evaluation-set:
       changes: {}
   - name: ""
     control-id: OSPS-AC-03
-    result: Passed
-    message: Branch protection rule prevents deletions
+    run: true
+    result:
+      state: Passed
+      message: Branch protection rule prevents deletions
     corrupted-state: false
     assessments:
     - requirement-id: OSPS-AC-03.01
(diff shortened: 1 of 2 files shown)


trumant:
I'd like to better understand how layer4 is expected to be used before an Assessment is run. I like the way you've modeled this in `cue` I'm just a skeptic that its a necessary change

jpower432:
> I'd like to better understand how layer4 is expected to be used before an Assessment is run. I like the way you've modeled this in `cue` I'm just a skeptic that its a necessary change

Thanks for the feedback! I'm glad you like the `cue` modeling approach. I appreciate your skepticism, and it's helpful to get your thoughts on this.

To better explain how Layer4 could be used before an Assessment is run, here are the two main uses cases I see this supporting:

1. Modeling Partial Executions: This update makes the outcome clearer for partial runs where an assessment may not have run due to failure, error, or inapplicability. For example, in the test data (I believe coming from a baseline validator scan), there are Assessment objects that were not run because they do not cover `Maturity Level 1` but there is results data anyway. That is captured through the `Not Run` Result, but the message is empty or might require placeholder data.

```
- name: ""
  control-id: OSPS-AC-04
  result: Not Run
  message: ""
  corrupted-state: false
  assessments:
  - requirement-id: OSPS-AC-04.01
    applicability:
    - Maturity Level 2
    - Maturity Level 3
    description: When a CI/CD task is executed with no permissions specified, the project's version control system MUST default to the lowest available permissions for all activities in the pipeline.
    result: Not Run
    message: ""
    steps:
    - github.com/revanite-io/pvtr-github-repo/evaluation_plans/osps/access_control.workflowDefaultReadPermissions
    steps-executed: 0
    start: 2025-08-22T16:02:00.000000000Z
    value: null
    changes: {}
```
2. Supporting Decoupled Test Plans: The Go library supports defining specific tests in a pre-run state. I believe it would be beneficial to support this type of lifecycle in other representations to enable use cases where the assessment orchestration tool is decoupled from the test definition.

With all that in mind, I'd like to understand your perspective better. I believe there's value in using Layer4 to model a test plan with introduction of procedures, but I'd love to know what you think. If you also see the value, do you think a test plan should be a completely separate object?

trumant:
I understand much better now, thanks Jen.

I like the concept of layer4 schema being useful for describing an evaluation plan/test plan. However, I think we would benefit from introducing a few new definitions/types that would help us provide greater clarity. Total rough draft follows (but perhaps something we can iterate on)

```
#EvaluationPlan: {
	author: #Contact
	plans: [#AssessmentPlan...]
}

#AssessmentPlan: {
	"control-id": string @go(ControlId)
	"assessment-requirements": [...#AssessmentRequirement] @go(AssessmentRequirements)
}

#AssessmentRequirement: {
	"requirement-id": string @go(RequirementId)
	procedures: [...#AssessmentProcedure] @go(Procedures)
}
```

Ideally we can reuse some definitions between EvaluationPlans and EvaluationResults. This would also open up the ability for the EvaluationResults to contain or otherwise reference the EvaluationPlan data. 

Going one step further, I think this opens up the ability to move applicability out of EvaluationResults and deal with it better in EvaluationPlan. I could see EvaluationPlan defining the applicability category values and when an evaluator takes that plan as input and produces an EvaluationResult the result only contains the applicable results

trumant:
> partial runs where an assessment may not have run due to failure, error, or inapplicability

As I said above, I'd prefer to see inapplicable controls omitted entirely from the results data payload.

So, if we adopt the idea of building some new types for EvaluationPlan, I think inapplicability goes away and then all assessments, etc in an EvaluationResult are always `run = true` and the result can communicate error, etc.

jpower432:
I really like where you're going with this draft! I have a couple of thoughts on this, and I've included a revised rough draft below that reflects them:

1.  We might want to tweak the names to avoid a collision with `AssessmentRequirement` in Layer 2.
2.  If I'm understanding correctly, your proposal implies that `AssessmentProcedure` would not be the parent of `AssessmentStep` as we discussed. Instead, `AssessmentSteps` would be recorded as an outcome of an object optionally linked to a procedure, reducing breaking changes in Layer 4 results.

Did I understand correctly?

```
#EvaluationPlan: {
	author: #Contact
	plans: [#AssessmentPlan...]
}

#AssessmentPlan: {
	"control-id": string @go(ControlId)
	"assessment": [...#Assessment] @go(Assessment)
}

#Assessment: {
	"requirement-id": string @go(RequirementId)
	procedures: [...#AssessmentProcedure] @go(Procedures)
}
```
_Note: Removed some fields for brevity_
```
#EvaluationResults: {
	...
}

#ControlEvaluation: {
	name:              string
	"control-id":      string @go(ControlId)
	assessment-logs: [...#AssessmentLog]
}

#AssessmentLog: {
	"requirement-id": string @go(RequirementId)
    "procedure-id"?: string @go(ProcedureId)
	description: string
	result:      #Result
	message:     string
	steps: [...#AssessmentStep]
	"start":           #Datetime
	"end"?:            #Datetime
	recommendation?: string
}
```

> Going one step further, I think this opens up the ability to move applicability out of EvaluationResults and deal with it better in EvaluationPlan. I could see EvaluationPlan defining the applicability category values and when an evaluator takes that plan as input and produces an EvaluationResult the result only contains the applicable results

I strongly agree &mdash; right now this type of logic is intertwined with the results. If we introduce an `EvaluationPlan` object as a separate artifact, I would expect a result for every applicable assessment.

jpower432:
I added a separate [PR](https://github.com/ossf/gemara/pull/117) to explore the `EvaluationPlan` to keep the changes in this PR scoped to the `Results`.

jpower432:
Closing this PR as I think this proposal is obselete.

Issue #91, opened 2025-08-14
Reduce CUE duplication


eddie-knight:
I feel like we want to avoid duplicating types across the schema and go packages.

Perhaps there is a way to define each layer schema as a collection of cue files where one or more of those files may contain types like those one which is reused across multiple layers

_Originally posted by @trumant in https://github.com/revanite-io/gemara/pull/90#discussion_r2275167216_

jpower432:
I am working through this issue currently defining a CUE module and layer level packages.

PR #117, opened 2025-09-02
feat!: adds `EvaluationPlan` to layer 4 schema

### LINKED ISSUE #73: Support mapping declarative policy identifiers or locations to `layer4.Assessments` as steps

jpower432:
## What Changed?

Introduce a new top level object to describe a test plan including `AssessmentProcedure`.

**Usage**: To support a human-led assessment, a test plan can described in an `EvaluationPlan` and a specific testing procedure can be described for a requirement.

Initially proposing, instead of duplicating this information in the `EvaluationResults`, the procedure can optionally be linked to an outcome of an `Assessment`.

Related to https://github.com/ossf/gemara/pull/113#issuecomment-3217411008
Supersedes #83

:warning: The breaking change here is for naming: `Assessment` -> `AssessmentLog`, but I'm open to other ideas on the naming.

### DIFF
--- a/layer4/evaluation_plan.go
+++ b/layer4/evaluation_plan.go
@@ -1,63 +1,63 @@
 package layer4
 
 type EvaluationPlan struct {
-	Metadata Metadata `json:"metadata"`
+	Metadata	Metadata	`json:"metadata" yaml:"metadata"`
 
-	Plans []AssessmentPlan `json:"plans"`
+	Plans	[]AssessmentPlan	`json:"plans" yaml:"plans"`
 }
 
 type Metadata struct {
-	Id string `json:"id"`
+	Id	string	`json:"id" yaml:"id"`
 
-	Version string `json:"version,omitempty"`
+	Version	string	`json:"version,omitempty" yaml:"version,omitempty"`
 
-	Author Contact `json:"author"`
+	Author	Contact	`json:"author" yaml:"author"`
 }
 
 type Contact struct {
 	// The contact person's name.
-	Name string `json:"name"`
+	Name	string	`json:"name" yaml:"name"`
 
 	// Indicates whether this admin is the first point of contact for inquiries. Only one entry should be marked as primary.
-	Primary bool `json:"primary"`
+	Primary	bool	`json:"primary" yaml:"primary"`
 
 	// The entity with which the contact is affiliated, such as a school or employer.
-	Affiliation *string `json:"affiliation,omitempty"`
+	Affiliation	*string	`json:"affiliation,omitempty" yaml:"affiliation,omitempty"`
 
 	// A preferred email address to reach the contact.
-	Email *string `json:"email,omitempty"`
+	Email	*string	`json:"email,omitempty" yaml:"email,omitempty"`
 
 	// A social media handle or profile for the contact.
-	Social *string `json:"social,omitempty"`
+	Social	*string	`json:"social,omitempty" yaml:"social,omitempty"`
 }
 
 // AssessmentPlan defines all testing procedures for a control id.
 type AssessmentPlan struct {
-	ControlId string `json:"control-id"`
+	ControlId	string	`json:"control-id" yaml:"control-id"`
 
-	Assessments []Assessment `json:"assessments"`
+	Assessments	[]Assessment	`json:"assessments" yaml:"assessments"`
 }
 
 // AssessmentPlan defines all testing procedures for a requirement.
 type Assessment struct {
 	// RequirementID is the unique identifier for the requirement being tested
-	RequirementId string `json:"requirement-id"`
+	RequirementId	string	`json:"requirement-id" yaml:"requirement-id"`
 
 	// Procedures defined possible testing procedures to evaluate the requirement.
-	Procedures []AssessmentProcedure `json:"procedures"`
+	Procedures	[]AssessmentProcedure	`json:"procedures" yaml:"procedures"`
 }
 
 // AssessmentProcedure describes a testing procedure for evaluating a Layer 2 control requirement.
 type AssessmentProcedure struct {
 	// Id uniquely identifies the assessment procedure being executed
-	Id string `json:"id"`
+	Id	string	`json:"id" yaml:"id"`
 
 	// Name provides a summary of the procedure
-	Name string `json:"name"`
+	Name	string	`json:"name" yaml:"name"`
 
 	// Description provides a detailed explanation of the procedure
-	Description string `json:"description"`
+	Description	string	`json:"description" yaml:"description"`
 
 	// Documentation provides a URL to documentation that describes how the assessment procedure evaluates the control requirement
-	Documentation string `json:"documentation,omitempty"`
+	Documentation	string	`json:"documentation,omitempty" yaml:"documentation,omitempty"`
 }
--- a/layer4/evaluation_plan.go
+++ b/layer4/evaluation_plan.go
@@ -1,63 +1,65 @@
 package layer4
 
+// EvaluationPlan defines how a set of Layer 4 controls are to be evaluated.
 type EvaluationPlan struct {
-	Metadata Metadata `json:"metadata" yaml:"metadata"`
+	Metadata	Metadata	`json:"metadata" yaml:"metadata"`
 
-	Plans []AssessmentPlan `json:"plans" yaml:"plans"`
+	Plans	[]AssessmentPlan	`json:"plans" yaml:"plans"`
 }
 
+// Metadata contains metadata about the evaluation plan.
 type Metadata struct {
-	Id string `json:"id" yaml:"id"`
+	Id	string	`json:"id" yaml:"id"`
 
-	Version string `json:"version,omitempty" yaml:"version,omitempty"`
+	Version	string	`json:"version,omitempty" yaml:"version,omitempty"`
 
-	Author Contact `json:"author" yaml:"author"`
+	Author	Contact	`json:"author" yaml:"author"`
 }
 
 type Contact struct {
 	// The contact person's name.
-	Name string `json:"name" yaml:"name"`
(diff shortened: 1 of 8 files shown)


eddie-knight:
I like this. I don't _think_ it's going to cause any unintended consequences... but it would be a breaking change, so I'd like to test it out locally with a branch of Privateer and the baseline plugin to kick the tires before we lock it in. It might be a couple days before I can knock that out.

trumant:
I like the direction is going, nice work @jpower432 

Like @eddie-knight , I want to dive a bit deeper here to kick the tires, but at first glance this looks really promising

trumant:
@jpower432 I've raised https://github.com/jpower432/sci/pull/10 with some proposed schema changes that I believe will help tighten up this change.

layer4/evaluation_plan.go
```diff
@@ -0,0 +1,63 @@
+package layer4
```
trumant, line 1: Any reason not to generate this file from the `cue` schema?

schemas/layer-4.cue
```diff
@@ -2,22 +2,29 @@ package schemas
...
 import "time"
 
+// EvaluationResults contains the results of evaluating a set of Layer 4 controls.
 #EvaluationResults: {
 	"evaluation-set": [#ControlEvaluation, ...#ControlEvaluation] @go(EvaluationSet)
 	...
 }
 
+// ControlEvaluation contains the results of evaluating a single Layer 4 control.
+// TODO: are all control requirements guaranteed to be evaluated at once, or does applicability influence this?
```
trumant, line 12: @jpower432 can you clarify where applicability lands in all this?

schemas/layer-4.cue
```diff
@@ -45,3 +52,57 @@ import "time"
...
+
+// AssessmentPlan defines all testing procedures for a control id.
+#AssessmentPlan: {
+	"control-id": string @go(ControlId)
+	"assessments": [...#Assessment] @go(Assessments)
+}
+
+// Assessment defines all testing procedures for a requirement.
+#Assessment: {
+	// RequirementID is the unique identifier for the requirement being tested.
```
trumant:
```suggestion
	// RequirementId is the unique identifier for the requirement being tested.
```

schemas/layer-4.cue
```diff
@@ -45,3 +52,57 @@ import "time"
...
+	// The contact person's name.
+	name: string
+	// Indicates whether this admin is the first point of contact for inquiries. Only one entry should be marked as primary.
+	primary: bool
+	// The entity with which the contact is affiliated, such as a school or employer.
+	affiliation?: string @go(Affiliation,type=*string)
+	// A preferred email address to reach the contact.
+	email?: #Email @go(Email,type=*Email)
+	// A social media handle or profile for the contact.
+	social?: string @go(Social,type=*string)
```
trumant, line 105: This seems superfluous, thoughts on removing it?

jpower432: This is duplicated from `layer3`. I thought by keeping them aligned, it would reduce breaking changes when we completed #91. Perhaps we should remove it from both? @eddie-knight Is there a defined use case for this field in `layer3`?

layer4/evaluation_plan.go
jpower432: I generated the types locally with `cue exp gengotypes` and copied over the subset here. I thought full cue generation for `layer4` was still blocked by the `changes` fields.

schemas/layer-4.cue
jpower432: I think `applicabilty` does influence whether the Assessment is run. From my perspective, the results should only contain the applicable results when a consumer runs `Evaluate`.

Currently, I believe is driven by the `layer4` relationship with `layer3`. With that said, maybe it makes sense to add `applicabilty` to the Assessment Plan so that `layer3` would not need to be referenced explicity and an engine could just execute the plan.

Below is what I am thinking the flow would be if we keep `EvaluationPlans` optional:
```
Layer3 Policy (with Applicability) --> Layer4 Plan (with Applicability) --> Layer4 Execution Engine --> Layer4 Results
Layer3 Policy (with Applicability) --> Layer4 Execution Engine --> Layer4 Results
```
WDYT?

schemas/layer-4.cue
eddie-knight: Inspiration for this came from Security Insights, where we see that some folks prefer to omit their email when putting a document out publicly — posting something like a github handle instead. It's a bit questionable whether that will ever create value in layer 3, but it's likely that layer 4 documents will be public attestations at some point.

schemas/layer-4.cue
trumant: > the results should only contain the applicable results when a consumer runs Evaluate

agreed!

> maybe it makes sense to add applicabilty to the Assessment Plan so that layer3 would not need to be referenced explicity and an engine could just execute the plan

To me, the process of producing the plan would consider applicability and only add AssessmentPlans for controls and Assessments for requirements that are applicable.

However, this can be resolved in a follow-on and we need not block these other changes.

trumant (approved):

schemas/layer-4.cue
evankanderson: Thinking about this from a verification perspective, is it useful for checkers / presentation layers to have the applicability information in the assessment? 

For example, if a project publishes content via GitHub, npm, and a website, a particular MFA assessment might only apply to GitHub but a separate assessment might be needed for npm.

(That's assuming:

1. All project resources and distribution channels are in scope for Gemara
2. We want to be able to use something like Gemara layer 4 in an in-toto envelope as an interchange format for assessment / attestation, e.g. for baseline

I'm dropping by thanks to Eddie's pointer; my primary interest is in building an ecosystem of assessment and presentation tools around a common assessment reporting format.)

schemas/layer-4.cue
```diff
@@ -2,22 +2,29 @@ package schemas
...
+// ControlEvaluation contains the results of evaluating a single Layer 4 control.
+// TODO: are all control requirements guaranteed to be evaluated at once, or does applicability influence this?
 #ControlEvaluation: {
 	name:              string
 	"control-id":      string @go(ControlId)
 	result:            #Result
 	message:           string
 	"corrupted-state": bool @go(CorruptedState)
-	assessments: [...#Assessment]
+	"assessment-logs": [...#AssessmentLog] @go(AssessmentLogs)
```
evankanderson, line 19: One small point for a separate PR (since it applies to a number of layer 4 fields).

The in-toto project discovered that `hyphen-separated` names are difficult to handle for a number of infix languages, including Python and JavaScript (maybe also Ruby, I'm not as familiar there).  Similarly, poking at these documents with `yq` will be more difficult (access as `.["assessment-log"][*]` rather than `.assessmentLog[*]`).

There may be other good reasons for this style, but I wanted to highlight the tradeoff before 1.0.

I'll open an issue for discussion.

evankanderson (commented):
Just dropping in, as I'd asked Eddie yesterday about the inclusion of `applicabilitly` to the layer 4 schema, which didn't seem well-documented in GitHub history.

schemas/layer-4.cue
jpower432: I'm interested in exploring this more. I think the possible differences in the way assessments could be performed was a driver for defining the `EvaluationPlan`.

> Thinking about this from a verification perspective, is it useful for checkers / presentation layers to have the applicability information in the assessment?

What I think we are exploring with regard to applicabilty here is defined in a higher layer in Layer 2 like `Maturity` level -- A project defines its target maturity in `layer3` to produce AssessmentPlan content and define what `ControlEvaluations` are in scope for exectution. For this case, having applicability defined in Layer 4 seems useful for checkers to understand the target at the time of the Assessment since Layer 4 is not explicity connected back to a Layer 3 policy version.

> For example, if a project publishes content via GitHub, npm, and a website, a particular MFA assessment might only apply to GitHub but a separate assessment might be needed for npm.

If I understand correctly, what you are suggesting in terms of applicabilty is more specific to the assessed project. 
I think this could be where `AssessmentProcedures` come in. For a single `Requirement`, you could define multiple procedures to defined ways an assessment could be performed possibly depending on the language, platform, or project tools. So `applicabilty` might be a way to enable procedure grouping or selection?

evankanderson: I'll admit that I haven't fully grokked Gemara (or thousands of other projects, but I commented on this one).  I'm also working off the names in the schema -- it seems like maybe the Go structs for layer 4 have richer comments than the schema definitions themselves.  (Though "applicability" doesn't reference that it's inherited from layer 3, so that may be an opportunity for improvement.)

It may be that having multiple AssessmentSteps (some of which may be no-ops) is the way to go.  So you might say "here is an assessment that any GitHub Repos have MFA" and "here is an assessment that any npm namespaces have MFA".  That seems reasonable to me, possibly alongside a "here is an assessment that the only types of resources the project has are {GitHub, npm, pypi, maven central, website}" to ensure that novel distribution methods are flagged.

jpower432: Are there any follow-on issues we can capture from this thread to allow the conversation to get marked as resolved?

eddie-knight (approved):

schemas/layer-4.cue
jpower432: Added https://github.com/ossf/gemara/issues/143 to capture this dicussion

PR #119, opened 2025-09-06
feat: Added layer2 loader LoadNestedCatalog

eddie-knight:
This allows for situations where a single file holds multiple catalogs — or data that isn't the catalog, such as the CCC release artifacts which look like this:

```yaml
catalog: # <...layer2.Catalog>
release-details: # <...CustomData>
```

This does the same thing as the rest of the loaders, but 1% different all the way through due to the data structure.

Theoretically we could make this support infinite nesting, but it would be more ugly than I want to deal with today... It'd need to accept a slice of strings and then loop through each string in the slice until it gets to the one we want to re-marshal and decode as a catalog. For our side it's not too complex, but I don't anticipate any use cases where someone would want to navigate that complexity.


layer2/loaders.go
```diff
@@ -93,6 +93,69 @@ func (c *Catalog) LoadFile(sourcePath string) error {
 	return nil
 }
 
+// Load a yaml file that contains a nested catalog
+// Only supports a single layer of nesting
```
trumant:
```suggestion
// LoadNestedCatalog retrieves a YAML file containing a nested catalog at the location fieldName from the remote location specified by sourcePath. The field specified by fieldName must be at the root level of the file.
```

layer2/loaders.go
```diff
@@ -93,6 +93,69 @@ func (c *Catalog) LoadFile(sourcePath string) error {
...
+		err = decoder.Decode(&yamlData)
+		if err != nil {
+			return fmt.Errorf("failed to decode YAML from URL: %v", err)
+		}
+	} else {
+		file, err := os.Open(sourcePath)
+		if err != nil {
+			return fmt.Errorf("error opening file: %w", err)
+		}
+		defer func() {
```
trumant: **Refactor Proposal for `LoadNestedCatalog`:**

```go
// LoadNestedCatalog loads a YAML file containing a nested catalog.
// Only supports a single layer of nesting.
// Accepts file URIs with the 'file:///' prefix.
// Throws an error if the URL is not https.
// TODO: Consider validating/sanitizing inputs to reduce injection risks.
func (c *Catalog) LoadNestedCatalog(sourcePath, fieldName string) error {
    if fieldName == "" {
        return fmt.Errorf("fieldName cannot be empty")
    }

    var yamlData map[string]interface{}

    parsedURL, err := url.Parse(sourcePath)
    if err != nil {
        return fmt.Errorf("failed to parse sourcePath: %w", err)
    }

    switch parsedURL.Scheme {
    case "http":
        // Reject http URLs for security reasons, only allow https/file.
        return fmt.Errorf("insecure http URLs are not allowed: %s", sourcePath)
    case "https":
        // TODO: Validate URLs to avoid SSRF and other external resource injection risks.
        resp, err := http.Get(sourcePath)
        if err != nil {
            return fmt.Errorf("failed to fetch URL: %v", err)
        }
        defer resp.Body.Close()
        if resp.StatusCode != http.StatusOK {
            return fmt.Errorf("failed to fetch URL; response status: %v", resp.Status)
        }
        decoder := yaml.NewDecoder(resp.Body)
        err = decoder.Decode(&yamlData)
        if err != nil {
            return fmt.Errorf("failed to decode YAML from URL: %v", err)
        }
    case "file":
        // TODO: Sanitize file paths to prevent directory traversal and unauthorized access.
        localPath := parsedURL.Path
        file, err := os.Open(localPath)
        if err != nil {
            return fmt.Errorf("error opening file: %w", err)
        }
        defer file.Close()
        decoder := yaml.NewDecoder(file)
        err = decoder.Decode(&yamlData)
        if err != nil {
            return fmt.Errorf("error decoding YAML: %w (%s)", err, localPath)
        }
    default:
        return fmt.Errorf("unsupported sourcePath scheme: %s", parsedURL.Scheme)
    }

    // TODO: Validate fieldName and the nested YAML content for injection risks.
    fieldData, exists := yamlData[fieldName]
    if !exists {
        return fmt.Errorf("field '%s' not found in YAML file", fieldName)
    }
    fieldYamlBytes, err := yaml.Marshal(fieldData)
    if err != nil {
        return fmt.Errorf("error marshaling field data to YAML: %w", err)
    }
    decoder := yaml.NewDecoder(strings.NewReader(string(fieldYamlBytes)))
    err = decoder.Decode(c)
    if err != nil {
        return fmt.Errorf("error decoding field '%s' into Catalog: %w", fieldName, err)
    }

    return nil
}
```

**Summary:**
- Uses Go comment conventions.
- Rejects `http://` URLs for security (only allows `https` and `file`).
- Handles file URIs with `file:///`.
- Includes `// TODO` comments for areas needing input validation and injection risk mitigation.

eddie-knight: Ran the unit tests and poked around with this a bit, seems to work as expected + has all the benefits you mention here. Could you past this into a suggestion block so that you get credit for the commit?

trumant: I'll do you one better and PR it against this branch

trumant: https://github.com/eddie-knight/gemara/pull/1/files

eddie-knight: I think that might be a larger changeset than I'm comfortable with in this PR... Happy to consider a larger refactor of the loaders, but it would be super helpful for me if we can get this one across the finish line first.

trumant (approved):

PR #154, opened 2025-10-01
fix: add implementation plan to layer 3 policy doc

jpower432:
The `ImplementationPlan` is defined in the `layer3` schema, but not connected to the `PolicyDocument`.

This PR adds the `implementation-plan` as an optional field to the policy document. Is it optional to avoid breaking changes.


eddie-knight (approved):
Good catch!

PR #156, opened 2025-10-02
fix: removed unstructured external references

eddie-knight:


jpower432:
Adding a note here from our discussion - we should update `prose` -> `text` to match Layer 2.

trumant (approved):

jpower432:
@eddie-knight PR to fix for the pipeline failure because if the OSCAL generation unit tests - https://github.com/eddie-knight/gemara/pull/2

eddie-knight:
Thanks Jenn!

Issue #157, opened 2025-10-05
Layer 2->3->4 schema relationships


eddie-knight:
After all the recent changes were applied, Layer 4 feels like it needs a big cleanup of how we're managing cross-document references. 

I think we're gonna need to remove some mappings to reduce redundancy and focus on the important pivots. And we need to add mapping references. And probably include some connection to the policy in the evaluation plan.

```mermaid


erDiagram

    %% === Core Document Structures ===
    EvaluationPlan {
        Metadata metadata
        AssessmentPlan[] plans
    }

    AssessmentPlan {
        string controlId "<- FK to Layer 2"
        Assessment[] assessments
    }


    Metadata {
        string id "PK"
        string version "optional"
        Evaluator evaluator
    }

    Evaluator {
        string name
        string uri "optional"
        string version "optional"
        Contact contact "optional"
    }

    ControlEvaluation {
        string name
        string controlId "<- FK to Layer 2"
        Result result
        string message
        AssessmentLog[] assessmentLogs
    }

    Assessment {
        string requirementId "<- FK to Layer 2"
        AssessmentProcedure[] procedures
    }

    AssessmentProcedure {
        string id "PK"
        string name
        string description
        string documentation "optional, URL"
    }

    AssessmentLog {
        string requirementId "<- FK to Layer 2"
        string procedureId "<- FK, optional"
        string[] applicability
        string description
        Result result
        string message
        string[] steps
        int stepsExecuted "optional"
        datetime start
        datetime end "optional"
        string recommendation "optional"
    }

    EvaluationLog {
        ControlEvaluation[] evaluations
        Metadata metadata "optional"
    }

    Result {
        string value "Not Run<br>Passed<br>Failed<br>Needs Review<br>Not Applicable<br>Unknown"
    }

    %% === Relationships ===


    EvaluationPlan }o--|{ AssessmentPlan : "contains"
    AssessmentPlan }o--|{ Assessment : "contains"
    Assessment }o--|{ AssessmentProcedure : "defines"

    EvaluationPlan |o--|| Metadata : "contains"
    EvaluationLog |o--|| Metadata : "contains"
    Metadata |o--|| Evaluator : "contains"

    EvaluationLog }o--|{ ControlEvaluation : "contains"

    ControlEvaluation }o--|{ AssessmentLog : "contains"

    AssessmentLog }o--|| AssessmentProcedure : "FK references"
    AssessmentLog ||--|| Result : "type"
[...]

PR #159, opened 2025-10-06
feat: Added mapping logic to layer 4

eddie-knight:
relates to #157

### DIFF
--- a/schemas/layer-4.cue
+++ b/schemas/layer-4.cue
@@ -18,9 +18,9 @@ import "time"
 
 // Metadata contains metadata about the Layer 4 evaluation plan and log.
 #Metadata: {
-	id:        string
-	version?:  string
-	author: #Author
+	id:       string
+	version?: string
+	author:   #Author
 	"mapping-references"?: [...#MappingReference] @go(MappingReferences) @yaml("mapping-references,omitempty")
 }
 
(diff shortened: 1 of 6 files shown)


schemas/layer-4.cue
```diff
@@ -18,46 +18,71 @@ import "time"
 
 // Metadata contains metadata about the Layer 4 evaluation plan and log.
 #Metadata: {
-	id:        string
-	version?:  string
-	evaluator: #Evaluator
+	id:       string
+	version?: string
+	author:   #Author
+	"mapping-references"?: [...#MappingReference] @go(MappingReferences) @yaml("mapping-references,omitempty")
```
jpower432, line 24: I know we have stuck to mapping references in the other layer schemas. Wondering if we should be more explict here. For instance, if a reference to a Layer 3 policy is required, should that getting its own dedicated field?
```suggestion
    "policy-reference": #MappingReference" @go(PolicyReference) @yaml(policy-reference)
	"mapping-references"?: [...#MappingReference] @go(MappingReferences) @yaml("mapping-references,omitempty")
```

eddie-knight: Mapping references are intended to be the location that a tool can refer to for information about an ID found elsewhere in the document. For example, if `ControlEvaluation.Requirement.ReferenceID` is `NIST_800_53` then we would look for `NIST_800_53` in `Metadata.MappingReferences` to get information about which version, where it was sourced from, etc.

jpower432: Ah, that's right. So to create that explicit relationship, we would use a `Mapping` instead. Policy reference would probably belong directly in the `EvaluationPlan` anyway. Resolving.

jpower432 (approved):

PR #162, opened 2025-10-09
fix: Enforce control and assessment reference documents match

evankanderson:
See discussion \[in Slack](https://openssf.slack.com/archives/C09A9PP765Q/p1760043482040019):

In particular:

* For any given control, all the assessments and procedures should refer to the same mapping reference document.  Omitting the reference document in the subordinate assessments and procedures defaults it to the same one as the control.
* This was tested with a sample document provided by Eddie in the thread.  The document failed with the following errors (once the mismatched header was removed).  Reported from `cue vet -c .\\schemas\\layer-4.cue .\\tempfile -d '#EvaluationLog'`:

   ```
   evaluations.0."assessment-logs".0.procedure."reference-id": conflicting values "OSPS-Baseline" and "":
        .\temp-openfga-values.yaml:10:23
        .\temp-openfga-values.yaml:17:25
   evaluations.1."assessment-logs".0.procedure."reference-id": conflicting values "OSPS-Baseline" and "":
        .\temp-openfga-values.yaml:45:23
        .\temp-openfga-values.yaml:52:25
   ...
   evaluations.12."assessment-logs".1.start: invalid value "" (does not satisfy time.Format("2006-01-02T15:04:05Z07:00")): error in call to time.Format: invalid time "":
       .\schemas\layer-4.cue:99:12
       .\schemas\layer-4.cue:88:9
       .\schemas\layer-4.cue:99:24
       .\temp-openfga-values.yaml:582:16
   ...
   ```

   The date errors seem to be pre-existing, and could be solved by changing `start` to `#Datetime | ""` or making `start` optional (possibly only for `Not Run` results).

### DIFF
--- a/schemas/layer-4.cue
+++ b/schemas/layer-4.cue
@@ -62,7 +62,6 @@ import "time"
 	// This formulation uses the control's reference if the assessment doesn't include a reference
 	"assessment-logs": [...{
 		requirement: "reference-id": (control."reference-id")
-		procedure: "reference-id":   (control."reference-id")
 	}] @go(AssessmentLogs,type=[]*AssessmentLog)
 }
 
--- a/schemas/layer-4.cue
+++ b/schemas/layer-4.cue
@@ -58,6 +58,12 @@ import "time"
 	message: string
 	control: #Mapping
 	"assessment-logs": [...#AssessmentLog] @go(AssessmentLogs,type=[]*AssessmentLog)
+	// Enforce that control reference and the assessments' references match
+	// This formulation uses the control's reference if the assessment doesn't include a reference
+	"assessment-logs": [...{
+		requirement: "reference-id": (control."reference-id")
+		procedure: "reference-id":   (control."reference-id")
+	}] @go(AssessmentLogs,type=[]*AssessmentLog)
 }
 
 // AssessmentLog contains the results of executing a single assessment procedure for a control requirement.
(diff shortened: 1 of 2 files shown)


schemas/layer-4.cue
```diff
@@ -58,6 +58,12 @@ import "time"
 	message: string
 	control: #Mapping
 	"assessment-logs": [...#AssessmentLog] @go(AssessmentLogs,type=[]*AssessmentLog)
+	// Enforce that control reference and the assessments' references match
+	// This formulation uses the control's reference if the assessment doesn't include a reference
+	"assessment-logs": [...{
+		requirement: "reference-id": (control."reference-id")
+		procedure: "reference-id":   (control."reference-id")
```
jpower432: Seeing this I am now thinking the `procedure` and `control` could potentionally come from different source documents. One from an `EvaluationPlan` and one from a source `Catalog`. I am thinking we might want to leave this one off for now.
```suggestion
```

jpower432 (commented):
Nice addition. I was able to test locally with the an example evaluation plan as well and it worked as expected. Left one thought on the `procedure` line.

schemas/layer-4.cue
eddie-knight: Can confirm these will be different

eddie-knight (approved):

eddie-knight:
Thanks for figuring this out @evankanderson, I didn't think this magic wand existed

evankanderson:
> Thanks for figuring this out @evankanderson, I didn't think this magic wand existed

I think this only works for `cue` (command and library); the generated go code allows a broader range of values.  This does document the expectation, though.

PR #171, opened 2025-11-02
feat: add PhysicalLocation support to SARIF output

zohayb23:
## Summary

Adds PhysicalLocation support to SARIF output to meet GitHub Code Scanning requirements. Each SARIF result now includes a PhysicalLocation with the artifact URI from Metadata.Author.Uri.

## Changes

- Add SARIF type structs: `PhysicalLocation`, `ArtifactLocation`, `Region`, and `Snippet`
- Extend `Location` struct to include optional `PhysicalLocation` field
- Update `ToSARIF()` to populate `PhysicalLocation` using `Metadata.Author.Uri` as the artifact URI
- Add tests to verify `PhysicalLocation` is included correctly
- Handle edge case when `Metadata.Author.Uri` is empty (nil PhysicalLocation)

## Testing

- All existing tests pass
- Added test to verify PhysicalLocation is included when URI is present
- Added test to verify PhysicalLocation is nil when URI is missing
- Backward compatible - LogicalLocations still present

## Related

Fixes missing PhysicalLocation required by GitHub Code Scanning for SARIF uploads.

### DIFF
--- a/layer4/evaluation_log.go
+++ b/layer4/evaluation_log.go
@@ -8,6 +8,11 @@ import (
 // ToSARIF converts the evaluation results into a SARIF document (v2.1.0).
 // Each AssessmentLog is emitted as a SARIF result. The rule id is derived from
 // the control id and requirement id.
+//
+// PhysicalLocation: Uses Metadata.Author.Uri as the artifact URI (typically a repository URL).
+// IMPORTANT: GitHub Code Scanning may require file paths (e.g., "README.md") instead of repository URLs.
+// If testing shows repository URLs don't work with GitHub Code Scanning, consider implementing
+// an optional artifactURI parameter (see Option A implementation below).
 func (e EvaluationLog) ToSARIF() ([]byte, error) {
 	report := &SarifReport{
 		Schema:  "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/123e95847b13fbdd4cbe2120fa5e33355d4a042b/Schemata/sarif-schema-2.1.0.json",
@@ -50,16 +55,37 @@ func (e EvaluationLog) ToSARIF() ([]byte, error) {
 				msg = log.Description
 			}
 
+			// Build PhysicalLocation using repository URI from metadata
+			// For repository-level assessments, use Metadata.Author.Uri as the artifact URI
+			// Region is left nil as we don't have file-specific line/column data
+			//
+			// NOTE: This uses Metadata.Author.Uri (repository URL like "https://github.com/ossf/gemera")
+			// GitHub Code Scanning may require file paths (e.g., "README.md") instead.
+			// TODO: Test with GitHub Code Scanning. If repository URLs don't work, implement
+			// Option A: Add optional artifactURI parameter to ToSARIF() method signature.
+			var physicalLocation *PhysicalLocation
+			if e.Metadata.Author.Uri != "" {
+				physicalLocation = &PhysicalLocation{
+					ArtifactLocation: ArtifactLocation{
+						URI: e.Metadata.Author.Uri,
+					},
+					// Region left nil - no line/column data available for repository-level assessments
+				}
+			}
+
+			location := Location{
+				PhysicalLocation: physicalLocation,
+				LogicalLocations: []LogicalLocation{
+					{FullyQualifiedName: ruleID},
+				},
+			}
+
 			result := ResultEntry{
 				RuleID:  ruleID,
 				Level:   level,
 				Message: Message{Text: msg},
 				Locations: []Location{
-					{
-						LogicalLocations: []LogicalLocation{
-							{FullyQualifiedName: ruleID},
-						},
-					},
+					location,
 				},
 			}
 			run.Results = append(run.Results, result)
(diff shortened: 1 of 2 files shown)


layer4/evaluation_log.go
```diff
@@ -50,16 +54,32 @@ func (e EvaluationLog) ToSARIF() ([]byte, error) {
 				msg = log.Description
 			}
 
+			// Build PhysicalLocation using repository URI from metadata.
```
jpower432: Wondering if the comments can be consolidated here. `Region` being nil is explained in few places with similar language.

eddie-knight: I'm also curious about `physicalLocation` compared to `logicalLocation`

https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790923

eddie-knight: Also, I'm curious whether we should be putting the code path from the step here

jpower432: I agree. For `PhysicalLocation`, if the resource being assessed is what we want to use, perhaps it could be passed as a parameter to `ToSarif()` since that information is not provided in the `EvaluationLog`. Using the `AssessmentStep` function address for `LogicalLocation` makes sense to me since that step would be the originator for the result.

zohayb23: @eddie-knight @jpower432 Thanks for the feedback! 

**PhysicalLocation vs LogicalLocation**: PhysicalLocation identifies the artifact (file/repo) where the result was found. LogicalLocation identifies the assessment step that produced it. I've clarified this in the docs.

**Code path from step**: I've implemented both suggestions:
- Added optional `artifactURI` parameter for PhysicalLocation (allows passing file paths like "README.md")
- LogicalLocation now uses AssessmentStep function address (`log.Steps[0].String()`) since the step is the originator

Both changes are backward compatible. Updated the code and tests accordingly.

layer4/evaluation_log.go
```diff
@@ -9,10 +9,15 @@ import (
...
-func (e EvaluationLog) ToSARIF() ([]byte, error) {
+// Parameters:
+//   - artifactURI (optional): File path or URI for PhysicalLocation.artifactLocation.uri.
+//     If not provided or empty, falls back to Metadata.Author.Uri (repository URL).
+//     For GitHub Code Scanning, typically use a file path like "README.md".
+//
+// PhysicalLocation identifies the artifact (file/repository) where the result was found.
+// LogicalLocation identifies the logical component (assessment step) that produced the result.
+// Region is left nil as we don't have file-specific line/column data.
+func (e EvaluationLog) ToSARIF(artifactURI ...string) ([]byte, error) {
```
eddie-knight: Why's this take one or more strings? I don't notice any place below where the additional values beyond the first are used

zohayb23: You're right, I used variadic parameters to make it optional, but since we only use the first value, it's misleading.

layer4/evaluation_log.go
```diff
@@ -50,16 +59,43 @@ func (e EvaluationLog) ToSARIF() ([]byte, error) {
 				msg = log.Description
 			}
 
+			// Determine artifact URI: use provided parameter, fallback to Metadata.Author.Uri
+			var uri string
+			if artifactURI != "" {
+				uri = artifactURI
+			} else if e.Metadata.Author.Uri != "" {
```
jpower432: I'm curious about this fallback behavior here. The `Author.Uri` is being used as the `ToolComponent` here as well (the entity that produced the log). Is there an example where this field is being used as the resource/repository URL?

layer4/evaluation_log.go
jpower432: This is more of a note for posterity, but since we are adding a physical location for every processed `AssessmentLog` it might be useful to re-examine @trumant's ideas about evidence/defined inputs for Layer 4 Evaluations in https://github.com/ossf/gemara/issues/23. I think that could make results like this more actionable.

layer4/evaluation_log.go
zohayb23: @jpower432 You're right - `Author.Uri` represents the tool/evaluator, not the resource being assessed. Using it as a fallback for PhysicalLocation is incorrect.

I'll remove the fallback - if `artifactURI` is empty, PhysicalLocation will be nil. This requires callers to explicitly provide the resource URI when needed.

layer4/evaluation_log.go
```diff
@@ -50,16 +59,38 @@ func (e EvaluationLog) ToSARIF() ([]byte, error) {
 				msg = log.Description
 			}
 
+			// PhysicalLocation: use provided artifactURI if available
+			// Note: Metadata.Author.Uri represents the tool/evaluator, not the resource being assessed,
+			// so we don't use it as a fallback for PhysicalLocation.
```
eddie-knight:
```suggestion
```

layer4/evaluation_log.go
```diff
@@ -50,16 +59,38 @@ func (e EvaluationLog) ToSARIF() ([]byte, error) {
...
+					},
+					// Region left nil - no line/column data available
+				}
+			}
+
+			// Use AssessmentStep function address for LogicalLocation (the step is the originator)
+			logicalLocationName := ruleID
+			if len(log.Steps) > 0 && log.Steps[0] != nil {
+				logicalLocationName = log.Steps[0].String()
+			}
```
eddie-knight: Ah. This is a sticky bit... My suggestion to put the step location here failed to consider that the location is being provided for the entire evaluation.

Perhaps we should give the location of the last step that was run?

zohayb23: That makes sense. Changes Committed.

layer4/evaluation_log.go
zohayb23: That's a good shout @jpower432. Looking into #23 now. I think this could be implemented in a future PR to enhance PhysicalLocation with actual evidence from each assessment, rather than using a generic artifactURI. Appreciate the feedback!.

eddie-knight (approved):

PR #182, opened 2025-11-13
fix: Set a default artifact location URI value

eddie-knight:
This resolves the error `locationFromSarifResult: expected a physical location` when artifactURI is not provided


jpower432 (approved):
LGTM

PR #200, opened 2025-12-02
feat!: adds unified package and reorganize schemas

### LINKED ISSUE #91: Reduce CUE duplication

jpower432:
## Description

This PR accomplishes the following:

-   Consolidating layer-based packages (layer1–layer4) into a unified structure
-   Adding common types
-   Reorganizing files and consolidating generated types
-   Updating CUE schemas
-   Adding documentation (ADR-0001)

### Schema Changes Made

- [ ] No schema changes
- [X] Layer 1 schema (`schemas/layer-1.cue`) changes
- [X] Layer 2 schema (`schemas/layer-2.cue`) changes
- [X] Layer 3 schema (`schemas/layer-3.cue`) changes
- [X] Layer 4 schema (`schemas/layer-4.cue`) changes

### Schema Change Details

Schema changed were made to consolidate common types like `Metadata`, `Mapping`, and `Contact`.
This is a breaking changes because the Metadata is updated to provide a common base for all top-level objects.

## Testing

- [X] Manual testing performed
- [X] Test data updated (if applicable)

## Related Issues

Closes #91 

## Reviewer Hints

In the second commit and an ADR documenting this decision and the tradeoffs was added


schemas/mapping.cue
```diff
@@ -0,0 +1,37 @@
...
+	"entry-id":      string @go(EntryId)
+	remarks?:        string
+}
+
+// MappingEntry represents a single entry within a mapping
+#MappingEntry: {
+	"reference-id": string @go(ReferenceId)
+	strength:       int & >=1 & <=10
+	remarks?:       string
+}
```
eddie-knight: Not sure if this is extremely clear as it stands. Perhaps we can find something to help the name communicate the purpose... maybe `Mapping` becomes `MultipleMapping` while `EntryMapping` becomes `SingleMapping` ...or something along those lines

schemas/base.cue
```diff
@@ -0,0 +1,54 @@
...
+	// Description provides additional context about the actor.
+	description?: string
+	// Uri provides a general URI for the actor information.
+	uri?: =~"^https?://[^\\s]+$"
+	// Contact provides contact information for the actor.
+	contact?: #Contact @go(Contact)
+}
+
+// ActorType specifies what entity is interacting in the workflow.
+#ActorType: "Human" | "Software" @go(-)
```
eddie-knight: Should we include something to note the use of AI-assistance, like `Hybrid` or `Software-Assisted`?

I know there's an argument that flagging AI generated content will eventually be a moot point, since it'll be used everywhere... but for highly-sensitive content, it feels important to allow users to note that not everything was hand-crafted.

eddie-knight (commented):
This looks good to me! Left a couple minor comments.

schemas/base.cue
jpower432: > Should we include something to note the use of AI-assistance, like `Hybrid` or `Software-Assisted`?
> 

I agree. I think this is a good add. Leaning toward "Software-Assisted" since it is understandable without knowing all of the options.

eddie-knight:
```suggestion
#ActorType: "Human" | "Software" | "Software-Assisted" @go(-)
```

eddie-knight (approved):

PR #207, opened 2025-12-06
feat: adds ConfidenceLevel to Assessment Log

### LINKED ISSUE #175: [Question] Should we support confidence levels in the Layer 4 `EvaluationLog`?
## Question

Should we allow tools that support Layer 4 to report a confidence level associated with the reported Result?

## Origination

`Would that be an output we want to add to Steps? ie,  Status=Passed;Confidence=10`
Idea from ORBIT WG Meeting (Nov6)

jpower432:
## Description

This PR as an optional `ConfidenceLevel` field to allow evaluators to add their confidence in a specific result.

### Schema Changes Made

- [ ] No schema changes
- [X] Layer 4 schema (`schemas/layer-4.cue`) changes

### Schema Change Details

`Layer 4 Assessment Log`
- Adds optional `ConfidencelLevel` field

## Testing

## Related Issues

Closes #175


assessment_log.go
```diff
@@ -80,6 +80,17 @@ func (a *AssessmentLog) Run(targetData interface{}) Result {
...
+// This method validates that an assessment has been run (Result != NotRun) before allowing
+// confidence to be set, as confidence only makes sense when there's an actual result.
+func (a *AssessmentLog) SetConfidenceLevel(confidenceLevel ConfidenceLevel) error {
+	if a.Result == NotRun {
+		return errors.New("cannot set confidence level before assessment has been run")
+	}
+	a.ConfidenceLevel = confidenceLevel
+	return nil
+}
+
```
eddie-knight: What if we tracked this on a per-step basis instead? When the result is issued, we could add a return value for confidence level.

For example, I might have 100% confidence if I can find a key indicator. But even if that key indicator is my only automation approach, I might have a low confidence simply if that key indicator isn't present. So the same test could be Pass = 10/10 confidence; Fail = 1/10 confidence.

OSPS Baseline values present in Security Insights are a good example of high confidence pass, low confidence failure.

jpower432: Thanks for the feedback @eddie-knight. The use case clarifies why the `Step` function has the context need to determine confidence. I'll update the step signature to return (Result, string, ConfidenceLevel), but I'll think I will need to tweak the loop logic so we don't overwrite the confidence with the last step's value (like `message`). I think it needs to aggregate like Result does or be tied to the Result that persists to the `Log`.

jpower432: @eddie-knight Pushed an update. Let me know what you think.

A couple points:
- I made `ConfidenceAggregation` stateful because "minimum confidence wins" (like `Result`) did not make sense to me in this case. I think additional steps could build or degrade confidence. I used a threshold-based approach, but we can talk about other options.
- Only results that actually affect the overall determined result are processed. If dropped during the aggregation (because of severity), it is not processed for confidence.

confidence_level.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+const (
+	// HighConfidenceThreshold is the minimum percentage (0.75 = 75%) of steps
+	// that must have High confidence for the aggregated result to be High.
+	HighConfidenceThreshold = 0.75
+	// MediumConfidenceThreshold is the minimum percentage (0.50 = 50%) of steps
+	// that must have Medium or High confidence for the aggregated result to be Medium.
+	MediumConfidenceThreshold = 0.50
+)
```
eddie-knight: I'm not convinced that counting passed steps is a good reflection of confidence.

For example: https://github.com/revanite-io/pvtr-github-repo/blob/main/evaluation_plans/evaluation-plans.go#L101-L106

The first two steps could pass, signifying that the test is applicable and should continue — but not providing any information relevant to the passing or failing of the assessment.

The only practical way I see to do this is to use the confidence score specified by the final step to run (ie, the last passing result or any other value)

eddie-knight (changes requested):
Only one issue I see. If we drop the aggregator concept, I'm quite happy with the rest of this  approach.

confidence_level.go
jpower432: That makes sense. If the convention is that all steps run prior to the final step (main assessment logic) will always be prereqs, I agree. Those steps shouldn't artificially inflate the confidence if the final step is where the real test logic is.

I also wonder about use cases here where we have two setup steps and two real checks like:
https://github.com/revanite-io/pvtr-github-repo/blob/04a90f9977571594ffc5d287b0bee69c209fab06/evaluation_plans/evaluation-plans.go#L107

If both the real checks pass and the first one was Low confidence and the second was High confidence. Would the overall confidence be High (Passing)?

jpower432: I removed the aggregation logic and added one tweak to your suggestion for the above use case. If the result is the same as the previous steps result, the confidence is only updated if it is lower than the current confidence to make sure we take the conservative approach for all steps that contribute the assessment result. There might be edge cases where that is an issue, like if an applicability check has a low confidence, but looking at the example, I wonder how likely that is. Let me know what you think.

jpower432:
As discussed in today's community meeting:

Updating this to use that last confidence level with the context that we want the author of the Assessment Steps to have fully control on how confidence builds as these steps are executed. This is also helpful because we don't have to account of the type or weight assign to the step. This can be done implicitly by the step implementer. A confidence in the step does not have to be isolated to the step outcome itself can adapt as each step is met and contribute to full confidence determination at the final step.

eddie-knight (approved):

PR #214, opened 2025-12-10
fix(docs): correct font backslant issue on mobile

sonupreetam:
## Description
Fixes backward italic text (backslant) caused by Cairo variable font's slant axis. Refactors `style.scss` for maintainability and switches to Google Fonts CDN, reducing file size while preserving all styling.

### Problem

The Cairo variable font (`Cairo-VariableFont_slnt,wght.ttf`) includes a `slnt` (slant) axis that ranges from 0° to -12° (backward slant), designed for Arabic RTL text. Without explicit control, browsers may render text with an unintended backward slant, causing all body text to appear as "backward italics."
This was most noticeable on mobile browsers (iOS Safari, Chrome mobile).
Also, tackle the problem of maintaining this in the long run.

### Solution

1. **Google Fonts CDN** — Replaced self-hosted fonts with CDN link
2. **SCSS Variables** — Single source of truth for colors, spacing, typography
3. **Removed Minima duplicates** — Print styles, wrapper, form focus, reduced motion
4. **Cleaned unused code** — Removed redundant variables

### Changes

| File | Change |
|------|--------|
| `docs/assets/css/style.scss` | Refactored with variables, removed `@font-face` blocks |
| `docs/_includes/custom-head.html` | Added Google Fonts CDN link |
| `docs/_config.yml` | With minima theme matching |
| `Gemfile.lock` | Updated version |

### Line Count

| Stage | Lines |
|-------|-------|
| Original | 394 |
| After variables + Google Fonts | 340 |
| After removing Minima duplicates | **287** |

### Code Review Notes

**Linter Warning (Line 14) shows a false positive linter error (`{ expected`). This is valid SCSS syntax for Jekyll's Sass pipeline - the CSS linter doesn't understand Sass imports. Can be safely ignored.

**Tablet Responsiveness:** The stylesheet relies on:
- Minima theme's built-in responsive breakpoints
- CSS Grid `auto-fit` for `.layer-grid` (adapts automatically)
- Flexible `max-width: 1200px` wrapper

### Testing

- [x] Desktop: Fonts load correctly via CDN
- [x] Mobile: No backward italic slant
- [x] All font weights render correctly
- [x] Brand colors applied consistently
- [x] `bundle exec jekyll build` succeeds

### Checklist

- [x] Clear title conforming to [Conventional Commits spec](https://www.conventionalcommits.org/)
- [x] DCO signoff on commits
- [x] No breaking changes

PR #220, opened 2025-12-12
fix: Revert #214

eddie-knight:
Reverts ossf/gemara#214

Apologies for merging these style changes without testing locally. I know better.

This is what the deployed site header looks like right now:

Screenshot 2025-12-12 at 10 18 33 AM


jpower432 (approved):
Thanks for catching this. It actually doesn't look like that on mobile so the I guess the issue shifted to desktop mode :smile:.

jpower432:
cc @sonupreetam

PR #231, opened 2025-12-19
docs: adds implementation versioning and maintenance documentation

### LINKED ISSUE #25: [design][release] schema versioning and release processes
# Overview

Consumers of the schemas want to understand how and when they change. They also want to reference the version of the schemas that their data conforms to when authoring and consuming said data.

# Next Steps

- [ ] Should we adopt SemVer and if so, at what scope? All layer schemas version together or each layer versions indepdently?
- [ ] How do schema version changes correspond to Github releases and versioning of the go module?
- [ ] Should we adopt some project guidelines or shared principles on the rate of release and rate of major and minor version releases?

jpower432:
## Description

This PR adds a versioning and maintenance process for the Gemara Implementation to the website.

## Rationale
The goal of this approach is to prioritize stability of the foundation over full completion, avoid releasing a stable monolithic solution or a perpetual draft. The proposal is that the first stable release would signify the stability of the concepts (Model & Lexicon) and create a stable, consistent path for schema implementation evolution.

- A stable schema means it has been implemented in at least one real world use case.
- An experimental schema requires feedback

### Schema Changes Made

- [X] No schema changes

Minor changes adding a new attribute

## Testing

## Related Issues

Closes #25

### DIFF
--- a/docs/implementation/index.md
+++ b/docs/implementation/index.md
@@ -103,7 +103,7 @@ Significant implementation changes are documented in [Architecture Decision Reco
 
 ## Versioning and Maintenance
 
-See the [Implementation Maintenance](maintenance.html) document for versioning and releases procedures.
+See the [Implementation Maintenance](maintenance) document for versioning and releases procedures.
 
 ## Relationship to Other Components
 
--- a/docs/implementation/index.md
+++ b/docs/implementation/index.md
@@ -52,9 +52,9 @@ go install cuelang.org/go/cmd/cue@latest
 cue vet ./your-controls.yaml ./schemas/layer-2.cue
 ```
 
-## Go Library
+## Go SDK
 
-Go libraries provide APIs for reading, writing, and manipulating Gemara documents.
+The Go SDK provides APIs for reading, writing, and manipulating Gemara documents.
 
 **[Go Package Reference →](https://pkg.go.dev/github.com/ossf/gemara)**
 
--- a/docs/implementation/index.md
+++ b/docs/implementation/index.md
@@ -75,7 +75,7 @@ import (
 )
 
 func main() {
-    catalog := &Catalog{}
+    catalog := &gemara.Catalog{}
     catalog, err := catalog.LoadFile("file://controls.yaml")
     if err != nil {
         panic(err)
@@ -100,6 +100,10 @@ See the [Contributing Guide](https://github.com/ossf/gemara/blob/main/CONTRIBUTI
 ## Architecture Decisions
 
 Significant implementation changes are documented in [Architecture Decision Records (ADRs)](../adr.html).
+
+## Versioning and Maintenance
+
+See the [Implementation Maintenance](maintenance.html) document for versioning and releases procedures.
 
 ## Relationship to Other Components
 
--- a/docs/implementation/maintenance.md
+++ b/docs/implementation/maintenance.md
@@ -36,7 +36,7 @@ Schemas follow this lifecycle: **Experimental** → **Stable** → **Deprecated*
 * Promoting a schema from Experimental to Stable triggers a **minor version increment** and involves a stabilization announcement in release notes, documentation updates, and tracking schema maturity.
 * Individual layers can be promoted independently (e.g., Layer 2 can be stable while Layer 1 remains in Experimental).
 * Once Stable, schemas **can still evolve** but maintain backward compatibility within major versions. Stable schemas allow **additive changes** such as new optional fields or new types.
-* Breaking changes to Stable schemas require a major version increment.
+* Breaking changes to Stable schemas require a major version increment. This should be avoided in all normal circumstances.
 * Stable schemas represent a long-term commitment and will continue to be supported.
 
 ### Deprecated Status
--- a/docs/implementation/maintenance.md
+++ b/docs/implementation/maintenance.md
@@ -16,6 +16,15 @@ Version increments follow SemVer:
 * **Minor version** increments for additive changes (new features, schema status promotions, or new fields/types in Stable schemas)
 * **Patch version** increments for bug fixes
 
+### Backward Compatible Changes to Stable Schemas
+
+Backward compatible changes to Stable schemas trigger **minor version increments**.
+
+Like:
+* Adding new optional fields to existing types
+* Adding new types
+* Adding new optional properties
+
 ## Release Process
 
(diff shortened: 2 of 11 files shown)


docs/implementation/maintenance.md
```diff
@@ -0,0 +1,62 @@
...
+* **Minor version** increments for additive changes (new features, schema status promotions, or new fields/types in Stable schemas)
+* **Patch version** increments for bug fixes
+
+## Release Process
+
+Changes are reviewed and tested before release. Git tags follow SemVer format (`v1.2.3`), and release notes document changes and migration paths.
+
+## Schema Promotion
+
+Schemas follow this lifecycle: **Experimental** → **Stable** → **Deprecated**. Status promotions trigger a **minor version increment**, as they are additive and do not break backward compatibility.
```
eddie-knight: Does this also apply to new fields within a stable schema?

docs/implementation/maintenance.md
```diff
@@ -0,0 +1,62 @@
...
+* Adding a new Experimental schema triggers a **minor version increment**.
+* Breaking changes and performance issues MAY occur.
+* Components **SHOULD NOT** be expected to be feature-complete.
+
+### Stable Status
+
+* Promoting a schema from Experimental to Stable triggers a **minor version increment** and involves a stabilization announcement in release notes, documentation updates, and tracking schema maturity.
+* Individual layers can be promoted independently (e.g., Layer 2 can be stable while Layer 1 remains in Experimental).
+* Once Stable, schemas **can still evolve** but maintain backward compatibility within major versions. Stable schemas allow **additive changes** such as new optional fields or new types.
+* Breaking changes to Stable schemas require a major version increment.
```
eddie-knight:
```suggestion
* Breaking changes to Stable schemas require a major version increment. This should be avoided in all normal circumstances.
```

docs/implementation/maintenance.md
```diff
@@ -0,0 +1,62 @@
...
+* Stable schemas represent a long-term commitment and will continue to be supported.
```
eddie-knight: What does that support look like?

docs/implementation/maintenance.md
jpower432: Yes. New backward-compatible fields in a stable schema should increment a minor version. Will add.

eddie-knight: Also, I mean to ask whether new fields can be marked as experimental or deprecated

jpower432: Oh interesting question. It might be useful to see if CUE has some functionality around that. I would think field-level deprecation would be useful. Experimental fields on a Stable schema seem counterintuitive to me.

I wonder if similarly to Security Insights, we might want to consider breaking changes at the field level in batches periodically.

Thinking through an example with the proposed lifecyle:

**Edited**:
```mermaid
graph TD
    %% Phase 1
    subgraph P1 [1. Prep Main]
        A[Main Branch: v1 Stable]
        A1[Action: Mark field deprecations]
        A --- A1
    end

    %% Phase 2
    subgraph P2 [2. The Split]
        B1[v1 Branch: Maintenance Mode]
        B2[Main Branch: v2 Experimental]
        note2[Note: Breaking changes allowed on Main]
        B1 --- B2
        B2 --- note2
    end

    %% Phase 3
    subgraph P3 [3. Promotion & Release]
        C1[v1 Branch: Deprecated]
        C2[Main Branch: v2 Stable]
        tag[Tag v2.0.0 Release]
        C1 --- C2
        C2 --- tag
    end

    P1 --> P2
    P2 --> P3
    
    style A fill:#d4edda,stroke:#28a745
    style B1 fill:#fff3cd,stroke:#ffc107
    style B2 fill:#f8d7da,stroke:#dc3545
    style C2 fill:#d4edda,stroke:#28a745
    style C1 fill:#e2e3e5,stroke:#6c757d
```

docs/implementation/maintenance.md
```diff
@@ -0,0 +1,81 @@
+---
+layout: page
+title: Maintenance
+---
+
+## Versioning Strategy
+
+Gemara follows [Semantic Versioning](https://semver.org/) (SemVer) with a **single release cycle** for both the Go module and CUE schemas.
+A single version tag (`v1.2.3`) applies to both the Go module (`github.com/ossf/gemara`) and the CUE schemas (`github.com/ossf/gemara/schemas`).
```
jpower432: Perhaps we should revisit this idea given the difference in completeness between the Go library and the schemas.

However, I think different release cycles for these implementations could be difficult in a monorepo structure when you consider Go package resolutions and what things could look post v1. In a monorepo, having the same release cycle makes more sense to me.

jpower432: Pushed an update with a proposal here. The practical implementation might need more discussion, but the vanity URL gives us some flexibility to figure that out.

**Edited**:

I believe we could also used the vanity URL for CUE as well.
So tags would be `schemas/v.X.Y.Z` for CUE and Go would be `vX.Y.Z`. If we moved the Go module to a new repo, it should be seamless because the module would also be at the root.

eddie-knight:
I like the new direction a lot. Still need to find the docs for the updated cue logic here though, looks interesting.

jpower432:
> Still need to find the docs for the updated cue logic here though, looks interesting.

This [doc](https://cuelang.org/docs/howto/conditionally-include-cue-files-using-build-attributes/) might be what you are looking for. The `status` properties were mainly to add metadata, but the build attributes could be helpful if we wanted to generate stable types and experimental types separately or only use stable constraints for verification:

Ex.
```
cue def -t stable  ./schemas/*.cue > stable.cue
cue exp gengotypes stable.cue
...
cue vet -t stable -d "#PolicyDocument" ./test-data/good-policy.yaml ./schemas/*.cue
```

docs/implementation/index.md
```diff
@@ -101,6 +101,10 @@ See the [Contributing Guide](https://github.com/ossf/gemara/blob/main/CONTRIBUTI
 
 Significant implementation changes are documented in [Architecture Decision Records (ADRs)](../adr.html).
 
+## Versioning and Maintenance
+
+See the [Implementation Maintenance](maintenance.html) document for versioning and releases procedures.
```
eddie-knight: nit: `.html` is uneccessary here

eddie-knight (approved):

PR #238, opened 2025-12-30
docs: added dark mode

eddie-knight:
## Description

This adds a dark mode feature to the site.

Debatably, it currently infers the user's preference from their system and loads that first. If we add this dark mode option, we may want to make it match other project websites... so we'll need to remove that logic to have it default to light mode until the user presses a button.

Screenshot 2025-12-30 at 12 25 38 PM

### Schema Changes Made

- [x] No schema changes

### Schema Change Details

```
```

## Testing

- [x] Manual testing performed

### DIFF
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -13,6 +13,10 @@
         </h1>
         <p class="site-tagline">GRC Engineering Model for Automated Risk Assessment</p>
       </div>
+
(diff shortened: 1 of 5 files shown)


jpower432 (approved):
Nice! Tried this on a few different browsers and the auto-detection works and the toggling works. I do like the auto-detection from a UX perspective.

PR #242, opened 2026-01-02
ci: Use official AIGF content for test data

eddie-knight:
### Schema Changes Made

- [x] No schema changes

### Schema Change Details

```
```

## Testing


eddie-knight:
Paused on this because I wanted to come up with an idea that doesn't involve setting explicit expectations in the test file by writing the same content in each.

I think it could be as simple as comparing the loaded object (before) to the oscal object (after)

PR #271, opened 2026-01-27
feat: applied feedback for GuidanceCatalog

eddie-knight:
## Description

Applies feedback and adds comments to Layer 1.

Does not add a schema for VectorCatalog.

### Schema Changes Made

- [ ] No schema changes
- [x] Layer 1 schema (`layer-1.cue`) changes
- [ ] Layer 2 schema (`layer-2.cue`) changes
- [ ] Layer 3 schema (`layer-3.cue`) changes
- [ ] Layer 5 schema (`layer-5.cue`) changes

### Schema Change Details

```
```

## Testing

### DIFF
--- a/layer-1.cue
+++ b/layer-1.cue
@@ -5,19 +5,30 @@ package schemas
 
 @go(gemara)
 
-#GuidanceDocument: {
-	title:           string
-	metadata:        #Metadata     @go(Metadata)
-	"document-type": #DocumentType @go(DocumentType) @yaml("document-type")
-	// Introductory text for the document to be used during rendering
+// GuidanceCatalog represents a concerted documentation effort to help bring about an optimal future without foreknowledge of the implementation details
+#GuidanceCatalog: {
+	// title describes the contents of this catalog at a glance
+	title:    string
+
+	// metadata provides detailed data about this catalog
+	metadata: #Metadata     @go(Metadata)
+
+	// type categorizes this document based on the intent of its contents
+	type: #GuidanceType @go(GuidanceType)
+
+	// front-matter provides introductory text for the document to be used during rendering
 	"front-matter"?: string @go(FrontMatter) @yaml("front-matter,omitempty")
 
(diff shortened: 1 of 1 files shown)


layer-1.cue
```diff
@@ -29,60 +40,84 @@ package schemas
...
+
+	// statements is a list of structural sub-requirements within a guideline
 	statements?: [...#Statement] @go(Statements)
 
+	// guideline-mappings documents the relationship between this guideline and external guidelines
 	"guideline-mappings"?: [...#MultiMapping] @go(GuidelineMappings) @yaml("guideline-mappings,omitempty")
-	// A list for associated key principle ids
+
+	// principle-mappings documents the relationship between this guideline and one or more principles
 	"principle-mappings"?: [...#MultiMapping] @go(PrincipleMappings) @yaml("principle-mappings,omitempty")
```
jpower432, line 91: With the amount of mappings we have here. I think it might be good to fully explore the utility of the principle mapping and whether we want to create a PrincipleCatalog.

layer-1.cue
```diff
@@ -29,60 +40,84 @@ package schemas
...
-	id:         string
-	title:      string
+	// id allows this entry to be referenced by other elements
+	id: string
+
+	// title describes the contents of this guideline
+	title: string
+
+	// objective is a unified statement of intent, which may encompass multiple situationally applicable statements
 	objective?: string
```
jpower432: In Controls this is required — I am wondering if we should do the same here. I don't know that statements and recommendations make as much sense without the objective tying it all together.

```suggestion
	objective: string
```

jpower432 (approved):
Approving with two comments that don't really apply to the actual changes being made here, but the schema as a whole.

layer-1.cue
eddie-knight: I think principles have always been debatable. I'd like to pin this for follow-up.

PR #275, opened 2026-01-27
docs: generate website pages for the schema documentation

eddie-knight:
## Description

booted up cursor to copy over the schema site builder from security-insights and adjust it to fit our multi-file schema

I came up with the idea of tracking the version in a file on security insights because I wanted a way to display the version on the website. We can tweak our release pipeline to only deploy the schema changes if the VERSION file does not contain -dev.

### Schema Changes Made

- [x] No schema changes
- [ ] Layer 1 schema (`layer-1.cue`) changes
- [ ] Layer 2 schema (`layer-2.cue`) changes
- [ ] Layer 3 schema (`layer-3.cue`) changes
- [ ] Layer 5 schema (`layer-5.cue`) changes

## Reviewer Hints

just build it locally and see if you catch anything I missed


cmd/cue2openapi/go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/gemaraproj/gemara/cmd/cue2openapi
+
+go 1.23.0
+
+require (
+	cuelang.org/go v0.7.0
+	gopkg.in/yaml.v3 v3.0.1
```
jpower432: This library is archived
```suggestion
	github.com/goccy/go-yaml
```

docs/_data/navigation.yml
```diff
@@ -8,4 +8,6 @@
   url: /adrs/
 - title: Community
   url: /community/
-  
\ No newline at end of file
+- title: Schema
```
jpower432, line 11: What if we move these under Implementation - https://gemara.openssf.org/implementation/ ? We are linking the schemas in the repositories right now. We could just link to these pages.

jpower432 (changes requested):
A couple thoughts:

- This adds a new field to the top navigation and sidebar. I think it would be good to sidebar to have it's own file because the dropdowns seem to set it apart from the top navigation. We if we had Quick Links that include the The Model and the Implementation.

- We already have a [location](https://github.com/gemaraproj/gemara/blob/main/docs/implementation/index.md) where we point to the schema in the repository and it would be ideal if they just linked out to the internal representations you made instead of duplicating with Schemas.

docs/_data/navigation.yml
eddie-knight: @jpower432 is it okay if we tackle this in the UI next steps that we discussed offline?

jpower432 (approved):
Approving this so we can complete incremental changes in smaller follow on PRs.

PR #407, opened 2026-05-11
chore: bump ospo-reusable-workflows release.yaml to v1.0.1

jmeridth:
## What

Pin the reusable release workflow to v1.0.0 (SHA 592067a69a43d2285f933753d89a7c9d51b96530). Add a Breaking Changes category to release-drafter.

## Why

v1.0.0 of ospo-reusable-workflows broadens the release trigger to include breaking, feature, vuln, and release labels and folds GoReleaser, container image build, attestation, and discussion creation into the reusable workflow itself. Surfacing breaking changes prominently in release notes aligns the changelog with the new label-based release triggers.

## Notes

- The outer label-filter `if:` block on the release job is removed because the v1.0 reusable workflow now handles label filtering internally.
- Trigger updated to `pull_request_target` so the workflow can push tags via `GITHUB_TOKEN`.
- The `publish-cue` downstream job is preserved untouched and still gated on `needs.release.outputs.full-tag`.

## Testing

After merging, trigger the release workflow via workflow_dispatch or merge a labeled PR.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -8,6 +8,9 @@ template: |
   See details of [all code changes](https://github.com/$OWNER/$REPOSITORY/compare/$PREVIOUS_TAG...v$RESOLVED_VERSION) since previous release
 
 categories:
+  - title: "💥 Breaking Changes"
+    labels:
+      - "breaking"
   - title: "🚀 Features"
     labels:
       - "feature"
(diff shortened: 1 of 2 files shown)


jmeridth:
@eddie-knight @jpower432 please add the release label to this PR. Please create the breaking label also.

jpower432 (approved):
LGTM. Reviewed the reusable workflow at `v1.0.1` and confirm label guard, and code is only checked out with `pull_request_target` when the PR is merged to main.

PR #420, opened 2026-06-12
docs: create ADR 0022 for Evidence in AssessmentLog

jpower432:
## Description

This PR adds an ADR for including evidence in `AssessmentLog` to enhance traceability of evaluations.

### Schema Changes Made

- [X] No schema changes
- [ ] Other

### Schema Change Details

```
```

## Testing

## Reviewer Hints

 ---

## Self-review checklist

### DIFF
--- a/docs/adrs/0022-evidence-on-assessment-log.md
+++ b/docs/adrs/0022-evidence-on-assessment-log.md
@@ -5,7 +5,7 @@ title: Add Evidence to AssessmentLog
 
 - **ADR:** 0022
 - **Proposal Author(s):** @jpower432, @eddie-knight
-- **Status:** Proposed
+- **Status:** Accepted
 
 ## Context
 
--- a/docs/adrs/0022-evidence-on-assessment-log.md
+++ b/docs/adrs/0022-evidence-on-assessment-log.md
@@ -1,0 +1,82 @@
+---
+layout: page
+title: Add Evidence to AssessmentLog
+---
+
+- **ADR:** 0022
+- **Proposal Author(s):** @jpower432, @eddie-knight
+- **Status:** Proposed
+
+## Context
+
+Gemara's measurement layers produce opinions, structured conclusions about the state of a target. An `EvaluationLog` is a tool's opinion: "based on what I observed, this control passes or fails." An `AuditLog` is an auditor's opinion: "based on the evidence I reviewed, this control is effective or not." Both opinions are rooted in evidence, the data that was actually observed that helped inform the actual conclusion. The tool sees it firsthand (an API response, a file, a config dump). The auditor sees it through the tool's record. Different proximity to the raw data, same root.
+
+Currently, the evidence root is invisible. An `AssessmentLog` records a pass/fail result with a message, but not "I consulted this specific resource at this time." The tool's opinion exists without a record of what it was rooted in, limiting downstream consumers of the log.
+
+## Action
+
+Reshape the existing `#Evidence` type in the experimental `auditlog.cue` to serve both the evaluation layer (`AssessmentLog`) and the audit layer (`AuditLog`). 
+The `#Evidence` type captures what was cited to support an opinion for a specific activity, raw data for the tool, evaluation and enforcement artifacts for the auditor.
+
+```cue
+#Evidence: {
+    // id uniquely identifies this evidence
+    id?: string
+
+    // type categorizes the kind of evidence (open enum)
+    type: #EvidenceType
+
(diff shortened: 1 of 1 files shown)


docs/adrs/0022-evidence-on-assessment-log.md
```diff
@@ -0,0 +1,82 @@
+---
+layout: page
+title: Add Evidence to AssessmentLog
+---
+
+- **ADR:** 0022
+- **Proposal Author(s):** @jpower432, @eddie-knight
+- **Status:** Proposed
```
eddie-knight:
```suggestion
- **Status:** Accepted
```

docs/adrs/0022-evidence-on-assessment-log.md
```diff
@@ -0,0 +1,82 @@
...
+`#EvidenceType` remains an open enum. Recommended values include artifact types already known to Gemara (e.g., `EvaluationLog`, `EnforcementLog`) plus categories for common evidence forms like provenance.
+
+**Changes to the Assessment Log**:
+```cue
+#AssessmentLog: {
+    // ... existing fields ...
+
+    // evidence records the raw data cited to support this assessment's opinion
+    evidence?: [#Evidence, ...#Evidence]
+}
```
eddie-knight, line 56: FYI @vinayad @jmeridth This will be useful when added to the new AI assistance in Privateer. We can have one piece of evidence linking to the API calls, and a follow-on that logs the assistant interaction

eddie-knight (changes requested):

docs/adrs/0022-evidence-on-assessment-log.md
```diff
@@ -0,0 +1,82 @@
...
+    id?: string
+
+    // type categorizes the kind of evidence (open enum)
+    type: #EvidenceType
+
+    // collected is the timestamp when the evidence was gathered
+    "collected-at": #Datetime
+
+    // address identifies where the evidence can be found
+    address?: string
```
jpower432, line 33: One thing we want to drill down on. Should this be a string or an `ArtifactMapping`?

hbraswelrh (approved):
LGTM.

eddie-knight (approved):

docs/adrs/0022-evidence-on-assessment-log.md
hbraswelrh: @jpower432 This seems like a natural place to use `ArtifactMapping` since the address would be the reference to the evidence and `ArtifactMapping` itself is more self-explanatory. Also, the `ArtifactMapping` fields `reference-id` and `remarks` are both already strings. 

What would be the true benefit of using a string rather than the `ArtifactMapping` schema?

eddie-knight: I follow the intuition, but... I think that would add a layer of complexity we don't want, since it would then require mapping entries, which themselves require versioning.

```
// ArtifactMapping represents a mapping to an external artifact or artifact entry
#ArtifactMapping: {
	// reference-id identifies an element from a MappingReference in the artifact's metadata
	"reference-id": string @go(ReferenceId)


	// remarks is prose regarding the mapped artifact or the mapping relationship
	remarks?: string
}
```

I suppose this could be fine if we intend to strongly encourage that users strictly control any evidence that isn't inlined.

jpower432: Same thing I was thinking. It could create confusion because the evidence has an `id` and so does the `mapping-reference`. Same with description. I would be wondering what to put where.

eddie-knight: So in practice this would mean that an evaluator would store evidence in a second file which has an ID and entries with their own IDs. We'll need an artifact type for that (#EvidenceDocument) and then the evaluator would add the link to that document in its metadata, so that AssessmentLogs can link to the evidence entry by ID.

hbraswelrh: I get that. Would it make sense to require `id` then and just use the string for address? Or is there a specific reason for the `id` being optional on the `#Evidence`?

jpower432: Good catch. Either way, I think `id` should be mandatory.

hbraswelrh: Sounds good. Thanks

eddie-knight:
@jpower432 I'm merging this now so that I can put up a PR against it

Issue #422, opened 2026-06-13
Modeling AI-agent threats as a Gemara threat catalogue: capability granularity and a small slice to validate


eeee2345:
Hello. I maintain Agent Threat Rules (ATR), an open MIT-licensed detection ruleset for AI agent attacks: over 640 machine-readable rules across 10 attack categories, each already cross-referenced to OWASP Agentic and LLM, MITRE ATLAS, NIST AI RMF, EU AI Act, and ISO 42001.

I would like to explore expressing these AI-agent threats as a Gemara threat catalogue, and I want to get the modeling right before proposing anything substantial.

Two questions:

1. `capabilities` is required on a Threat. For agent-attack threats (for example tool-response injection, or cross-agent privilege escalation), what granularity do you intend for capabilities? The Capability Profiles work in #419 looks directly related. Is there guidance or an example I should follow so a catalogue fits that direction?

2. Would a small slice of 5 to 10 threats be the right way to validate the modeling first, before a fuller catalogue?

Happy to follow the ADR process. I am not looking to drop a full catalogue; I would rather align on the model first.

jpower432:
Hi @eeee2345. Welcome! Really glad to see interest in modeling AI agent threats with Gemara, and ATR's existing cross-references to OWASP, MITRE ATLAS, etc. would map well to Gemara's `#MappingDocument` artifacts too.

On question 1: One clarification -- Gemara defines the schema, but does not store the content. Your ATR catalog would live in your repository, validated against the Gemara schemas. However, we are working on tooling to help with distribution, specifically via OCI artfiacts 👉  https://github.com/gemaraproj/gemara-publish-action.

capabilities on a `#Threat` are a reference to entries in a separate `#CapabilityCatalog`. You'd author an AI agent `#CapabilityCatalog` describing what agent systems can do, things like tool invocation, cross-agent message passing, then reference those from your threats. A capability should describe a feature or function that creates the opportunity for the threat. FINOS Common Cloud Controls [catalogs](https://github.com/finos/common-cloud-controls/blob/main/catalogs/core/ccc/capabilities.yaml) are a good example to look at.

The Capability Profiles work in #419 is about how external CUE modules can extend `#Capability` with domain-specific fields. A plain [#CapabilityCatalog](https://github.com/gemaraproj/gemara/blob/main/capabilitycatalog.cue) can work today, but if you're interested in building an AI agent Capability profile, I'd be happy to work on that with you. An ADR would be the right vehicle for proposing it.

[...]

Issue #419, opened 2026-06-11
[Question] Should we support Capability Profiles?


jpower432:
## Motivation
  
  Organizations need to inventory their system capabilities in domain-specific terms. A Kubernetes team thinks in API resources and admission controllers. A Linux security team thinks in kernel capabilities and syscalls. An identity governance team thinks in the authorization surface their system exposes.

  Today, `#CapabilityCatalog` gives them `id`, `title`, `description`, and `group`, enough to name a capability, but not
  enough to describe it in the language of the domain.
  
  ### What is a Capability?
  
  A Gemara capability describes *what a system can do*, a feature, component, or object. Capabilities are neutral: they flow
  into threats (how the capability can be exploited), controls (how to mitigate), and assessments (how to verify). They are
  properties of the system, not of the users or policies that govern the system.
  
  For example:
  - "The system can pull container images by tag", a capability. It can be exploited (tag mutation attacks), controlled
  (require digest-based references), and assessed (verify no tag-only references in deployment configs).
  - "The system exposes a funds transfer API", also a capability. The system is capable of moving money between accounts. The
  threat is unauthorized access to that surface. The control is proper authorization enforcement.
  
  Profiles add domain-specific structure to these capabilities. A Kubernetes profile adds API group and admission controller
  fields. An identity profile adds the `action` and `resource` that describe the authorization surface the system exposes, what the system is capable of doing, not who is allowed to do it. That distinction matters: "the system exposes `transfer` on `bank-account`" is a capability; "Alice is authorized to transfer funds" is an entitlement, which is runtime policy and not a capability.
  
  ## Capability Profiles

  A profile is a CUE module that imports Gemara and refines `#Capability` via unification. No changes to the Gemara schema are
  required.

  ### GovOps snippet

**EDITED** to better match GovOps design

  ```cue
  package authorization

  import "github.com/gemaraproj/gemara@v1:gemara"
  
  #AuthorizationCapability: {
      gemara.#Capability
      action:              string
      resource:            string
      "risk-tier"?:        "critical" | "high" | "medium" | "low"
      "data-sensitivity"?: "public" | "internal" | "confidential" | "restricted"
  }
  
  #AuthorizationCapabilityCatalog: {
      gemara.#CapabilityCatalog
      capabilities: [#AuthorizationCapability, ...#AuthorizationCapability]
  }
  ```
  
  ### Kubernetes
  
  ```cue
  package kubernetes

[...]

PR #428, opened 2026-06-28
feat: add AI agent capability catalog and ATR category mappings

eeee2345:
Follows up gemaraproj/gemara#422. Adds a first slice modeling AI agent threats
with Gemara: an AI Agent CapabilityCatalog plus a MappingDocument relating Agent
Threat Rules (ATR) detection categories to those capabilities.

Per the discussion in #422, Gemara defines the schema and ATR hosts the content.
These example artifacts are authored by ATR and validated here against the Gemara
schemas so the structure can be reviewed and iterated.

What is included
- examples/ai-agent/ai-agent-capability-catalog.yaml: a CapabilityCatalog
  describing the agent features that create attack surface, including model
  inference over untrusted input, tool invocation, MCP and tool server
  connection, cross-agent message passing, skill and plugin loading, memory
  read and write, autonomous action, and privileged execution. Each entry
  describes a capability, not a threat, so threats can reference them.
- examples/ai-agent/atr-categories-to-capabilities-mapping.yaml: a
  MappingDocument relating the nine ATR detection categories to those
  capabilities with per-target strength, confidence, and rationale.
- test/schema_test.go: two positive validation cases for the artifacts.

Validation
- cue vet -d '#CapabilityCatalog' and -d '#MappingDocument' both pass.
- go test ./... passes, including the two new cases.

Open question for review: does this capability granularity match what you had
in mind, or would you scope finer or coarser? Happy to move the AI agent
Capability profile discussion (the #419 direction) into an ADR if useful.

### DIFF
--- a/examples/ai-agent/ai-agent-capability-catalog.yaml
+++ b/examples/ai-agent/ai-agent-capability-catalog.yaml
@@ -1,0 +1,143 @@
+metadata:
+  id: ATR-AI-AGENT-CAPABILITY-CATALOG
+  type: CapabilityCatalog
+  gemara-version: "1.1.0"
+  version: "0.1.0"
+  description: >
+    Capabilities of AI agent systems that create the opportunity for the
+    threats catalogued by Agent Threat Rules (ATR). Each capability describes
+    a feature or function an agent can perform, not a threat. Threats reference
+    these capabilities; ATR detection rule categories map to them in a companion
+    MappingDocument. Authored by the ATR project and validated against the
+    Gemara schemas; ATR hosts the content, Gemara defines the schema.
+  author:
+    id: atr
+    name: Adam Lin
+    type: Human
+    contact:
+      name: Adam Lin
+      affiliation: Agent Threat Rules
+      email: adam@agentthreatrule.org
+      social: eeee2345
+
+title: AI Agent Capability Catalog
+
+groups:
+  - id: model-interaction
+    title: Model Interaction
+    description: >
+      Capabilities by which an agent receives instructions and untrusted
+      content into a language model and acts on the model's output.
+  - id: tool-and-protocol
+    title: Tool and Protocol Access
+    description: >
+      Capabilities by which an agent invokes external tools and connects to
+      tool servers over protocols such as the Model Context Protocol (MCP).
+  - id: extensibility
+    title: Extensibility
+    description: >
+      Capabilities by which an agent loads and runs third-party extensions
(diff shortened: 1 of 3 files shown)


examples/ai-agent/ai-agent-capability-catalog.yaml
```diff
@@ -0,0 +1,143 @@
+metadata:
+  id: ATR-AI-AGENT-CAPABILITY-CATALOG
+  type: CapabilityCatalog
+  gemara-version: "1.1.0"
+  version: "0.1.0"
+  description: >
```
jpower432, line 6: @eeee2345 Capability set looks solid. I think the granularity hits the good level -- concrete enough to map distinct threat categories against, abstract enough to not couple to specific agent implementations.

examples/ai-agent/atr-categories-to-capabilities-mapping.yaml
```diff
@@ -0,0 +1,207 @@
+title: ATR Rule Categories to AI Agent Capabilities
+metadata:
+  id: ATR-CAP-MAP-001
+  version: "0.1.0"
+  type: MappingDocument
```
jpower432, line 5: The mapping document makes sense as a first slice. An alternative structure to consider would be capability→threat and threat→control through the inline mappings of `ThreatCatalogs` and `ControlCatalogs`. That gives you more expressiveness (which specific threats bridge a capability to a detection rule) at the cost of more artifacts to maintain.

jpower432 (approved):

jpower432:
@eddie-knight I will leave these conversations open for a bit in case you have additional feedback.

eeee2345:
@jpower432 thanks for the review, and apologies for the slow turnaround on my side.

On the alternative structure — you're right that routing capability→threat and threat→control through inline `ThreatCatalogs` / `ControlCatalogs` mappings is more expressive; naming *which* threat bridges a capability to a detection rule beats the flat mapping this PR ships. I'd suggest taking it as a follow-up rather than folding it in: the current slice is usable as-is, and the richer structure is worth doing deliberately — it's exactly the "more artifacts to maintain" cost you flagged, so it deserves its own change with the catalog wiring done properly.

If that split works for you and @eddie-knight has no further feedback, this is ready to merge — I'll open the threat→control inline-mapping follow-up against the merged base.

jpower432:
Sounds good as a follow-on @eeee2345. Will merge.

PR #433, opened 2026-07-08
docs: clarify that mapping-reference.id must match referenced artifact's metadata.id

### LINKED ISSUE #432: Document that mapping-reference.id must match the referenced artifact's metadata.id
## Summary

The CUE schema for `#MappingReference` allows any string for the `id` field, with no guidance on what value it should have. In practice, `mapping-reference.id` must match the referenced artifact's `metadata.id` for import resolution to work. This constraint should be documented in the schema definition. 

## Current schema (from `#MappingReference`)

The `id` field description says:
> id allows this entry to be referenced by other elements

This is accurate but incomplete — it does not mention that the value must also match the target artifact's `metadata.id`. @hbraswelrh suggested this schema-related documentation issue.

## Suggestion

Update the `id` field description in `#MappingReference` to state that it must match the referenced artifact's `metadata.id`. For example:

> id allows this entry to be referenced by other elements within this artifact. Must match the metadata.id of the referenced artifact for import resolution.

## Related

- gemaraproj/go-gemara#91 — cross-document validation to catch mismatches at the library level
- complytime/complypack#109 — complypack now surfaces a descriptive error when this mismatch occurs

sonupreetam:
## Summary

- Update the `id` field comment in `#MappingReference` to document the contract between `mapping-reference.id` and the referenced artifact's `metadata.id`

This is a **documentation-only** change. CUE comments don't affect validation,
and CUE cannot enforce cross-document constraints (each artifact is validated
independently). Runtime enforcement is handled separately in go-gemara via
`bundle.ValidateMappingReferences` (see gemaraproj/go-gemara#91).

## Change

Before:
```
// id allows this entry to be referenced by other elements
```

After:
```
// id allows this entry to be referenced by other elements within this artifact.
// When url is absent, should match the metadata.id of the referenced artifact
// to enable bundle-local import resolution.
```

The wording uses "should" rather than "must" because URL-bearing references
are resolved externally by the assembler — the ID match is only required
for bundle-local (URL-less) resolution.

## Related

- Closes #432
- gemaraproj/go-gemara#91 — runtime validation that enforces this contract for URL-less references
- complytime/complypack#109 — complypack error that motivated this documentation

## Test plan

- [x] `cue vet ./...` passes
- [x] `go test ./...` in `test/` passes

### DIFF
--- a/mapping_inline.cue
+++ b/mapping_inline.cue
@@ -7,7 +7,8 @@ package gemara
 
 // MappingReference represents a reference to an external document with full metadata.
 #MappingReference: {
-	// id allows this entry to be referenced by other elements
+	// id identifies this mapping reference within the artifact and, when url
+	// is absent, the referenced artifact's metadata.id.
 	id: string
 
 	// title describes the purpose of this mapping reference at a glance


jpower432:
@sonupreetam This one might just need a rebase now.

jpower432:
@sonupreetam Also thinking about this more, it seems like we are including SDK specific details in the schema documentation, which seems like it could be confusing.

sonupreetam:
Thank you @jpower432, you are right "bundle-local import resolution" seems is leaking SDK implementation here. How about below, amended the commit?
```
// id identifies this mapping reference within the artifact and, when url
// is absent, the referenced artifact's metadata.id.
```

jpower432 (approved):

PR #450, opened 2026-07-28
feat: add evidence mapping support to Evidence and AssessmentPlan

### LINKED ISSUE #417: Add Optional `#Evidence` type to `#AssessmentLog`
### Is your feature request related to a problem?

The evidence chain (EvaluationLog → EnforcementLog → AuditLog) traces across logs, but there is no way to record what specifically the evaluator examined within the target during an assessment. 

AssessmentLog records pass/fail with a message, but not *I consulted THIS specific resource at THIS time*. This means consumers like SARIF converters cannot populate file locations effectively, and auditors cannot verify what data sources the evaluator actually consulted without requesting supplementary evidence.

### Describe the solution you'd like

 Add a `#Provenance` type and optional field to `#AssessmentLog` in `evaluationlog.cue`:

Example
```
  #Provenance: {
      // address identifies what was consulted (file path, API endpoint, ARN, image ref, etc.)
      address:   string
      // collected is the timestamp when the data was acquired
      collected: #Datetime @go(Collected)
      // observed is the relevant content extracted
      observed?: string
  }
```
Three fields (two required, one optional), modality-neutral -- works for file scans, API checks, container scans, cloud config, physical audits, and interviews. 

On #AssessmentLog:
The field is optional and backward-compatible.
      `provenance?: [#Provenance, ...#Provenance] @go(Provenance,type=[]Provenance)`

### Describe alternatives you've considered

  Extensive naming exploration rejected 11 alternatives:

  - #Observation — collides with #ResultType: "Observation" enum value in auditlog.cue
  - #Subject — in OSCAL, subject means the component (inventory item instance) being assessed (equivalent to Gemara's target)
  - #Evidence — already exists on #AuditResult for macro-level log references
  - #Resource — already exists as a type (#Entity extension)
  - #Source — source field already exists on #Mapping in mappingdocument.cue
  - #Sample — implies statistical sampling methodology
  - #Examination — synonym/subset of Assessment and Evaluation

### Success Criteria

  - Criterion 1: #Provenance type and provenance? field on #AssessmentLog are added to evaluationlog.cue and pass CUE validation
  - Criterion 2: Existing EvaluationLog YAML documents validate without changes (backward-compatible)
  - Criterion 3: The backward compatibility check (test/compat_test.go) passes with the new field
  - Criterion 4: go-gemara can generate types from the updated schema and round-trip provenance data through YAML serialization

### Additional context

```
assessment-logs:
    - requirement:
        reference-id: osps-baseline
        entry-id: OSPS-AC-01.01
      description: "MFA enforcement check"
      result: Passed
      message: "MFA is enabled for all organization members"
      provenance:
        - address: "https://api.github.com/orgs/myorg/settings"
          collected: "2025-08-22T16:02:00Z"
          observed: '{"two_factor_requirement_enabled": true}'
        - address: "https://api.github.com/orgs/myorg/members?filter=2fa_disabled"
          collected: "2025-08-22T16:02:01Z"
          observed: "[]"
```
Rel #23

jpower432:
## Description

This PR adds an `EvidenceMapping` type. Details below.

### Schema Changes Made

- [ ] No schema changes
- [X] Audit Log schema (`auditlog.cue`) changes
- [X] Evaluation Log schema (`evaluationlog.cue`) changes
- [ ] Other

### Schema Change Details

#EvidenceMapping added to track evidence sources with:
- reference-id: link to mapping-reference in artifact metadata
- coordinate: specific location within source (line, JSON path, API endpoint)
- entry-id: structured entry identifier in Gemara artifacts
- digest: hash of evidence content at collection time
- remarks: prose notes about the reference

#Evidence now has optional origin field linking to #EvidenceMapping, allowing evidence to cite its source without repeating artifact details.

## Testing

- [X] Test data updated (if applicable)

## Related Issues

Closes #417 

## Reviewer Hints

 ---

## Self-review checklist


eddie-knight (dismissed):

jpower432:
Thanks for the approval @eddie-knight. Since this changes a stable schema, thought we might benefit from some extra reviews.

mapping_inline.cue
hbraswelrh, line 1: @jpower432 Would the `EvidenceMapping` every be an option or additional field on the EvaluationLog's [AssessmentLog.steps](https://github.com/gemaraproj/gemara/blob/093e9811918defb56a8ae41360d0f2f8a22647f8/evaluationlog.cue#L47)? This seems like a relevant area where additional context could optionally be added to the assessment step location.

hbraswelrh (dismissed):
LGTM. Left one question in the comments.

marcusburghardt:
:robot: I used AI assistance to evaluate this PR, test locally, confirm insights, and draft the feedback.

Nice work on this -- the evidence provenance gap is a real problem for auditability, and the `reference-id` indirection through `mapping-references` is a clean pattern that avoids repetition and stays consistent with `#ArtifactMapping` / `#EntryMapping`.

A few points for consideration before this lands as stable (since `mapping_inline.cue` carries `@status("stable")`):

### 1. Naming: `origin` vs alternatives

`origin` is concise, but in a git-centric open source ecosystem it strongly connotes "remote repository." When someone reads `origin: reference-id: github-api`, there is a moment of disambiguation. Have you considered `source` or `source-ref`? `source` reads more naturally for "where this evidence was collected from" and avoids the git overload. Not a blocker -- just worth a quick sanity check since renaming after stable is expensive.

### 2. `coordinate` ambiguity

The `coordinate` field is intentionally flexible, but consumers will receive values like `/orgs/foo/settings` (API path), `line 42` (file location), or `$.results[0].status` (JSONPath) with no way to distinguish the format programmatically. This works for human readers, but tools that need to render or cross-reference evidence locations would need to guess the format from context.

Two lightweight options that preserve simplicity:
- Add an optional `coordinate-type` field (`"api-path"`, `"json-pointer"`, `"file-location"`, etc.) so consumers can parse without guessing.
- Document a convention in the comment (e.g., "URI-style coordinates are recommended") even if not schema-enforced.

This could also be deferred to a follow-up if you want to keep the initial surface area small and see how adopters actually use `coordinate` in practice.

### 3. Single vs multiple origins

If a compliance finding is derived by correlating two sources (e.g., an API response and a config file), the singular `origin` field requires either splitting the finding into separate evidence entries or losing one of the source references. Has there been discussion about whether `origin` should be a list (`origins?: [#EvidenceMapping, ...]`)? The singular form is simpler and probably covers the majority of cases, but it would be good to confirm this is a deliberate choice rather than something that needs revisiting later.

### 4. `digest` format

The `digest` field has no format constraint. Since the test data uses `sha256:a1b2c3d4e5f6...`, a regex like `=~"^[a-z0-9]+:[a-f0-9]+$"` would enforce the `algorithm:hex` convention and align with OCI digest formats. This prevents malformed values from passing validation silently. If there are use cases for non-standard digest formats, documenting the expected format in the comment would help.

### 5. Schema lifecycle

`#EvidenceMapping` is being added to `mapping_inline.cue` which is `@status("stable")`. This means the type ships as stable on merge. If any of the points above (coordinate structure, singular vs list, digest format) might lead to field changes, it may be worth either addressing them now or adding the type in a separate file with `@status("experimental")` until the field set is validated by early adopters. Once stable, removing or restructuring fields becomes a breaking change.

---

None of these are blockers -- the design is sound for the current use cases and the backwards compatibility is clean. These are meant to help avoid future migration costs given the stable lifecycle status.

sonupreetam (approved):
Reviewed the schema changes, test data, cross-layer interactions and and alignment with meeting decisions. 

- For the schema design, the `coordinate` field for re-verification  and `digest` for verification are nice. 
- Apart from the concern @hbraswelrh  raised regarding whether `EvidenceMapping` could also be an option on `AssessmentLog.steps` for per-step source attribution, or maybe we are missing something. 

Everything LGTM.

jpower432:
Thanks for the feedback @marcusburghardt. Addressing one of the points below.

> `#EvidenceMapping` is being added to `mapping_inline.cue` which is `@status("stable")`. This means the type ships as stable on merge. If any of the points above (coordinate structure, singular vs list, digest format) might lead to field changes, it may be worth either addressing them now or adding the type in a separate file with `@status("experimental")` until the field set is validated by early adopters. Once stable, removing or restructuring fields becomes a breaking change.

This is being used in the `#EvaluationLog` type which is stable. We have a [versioning rule ](https://gemara.openssf.org/schema/)that says stable objects cannot reference experimental objects.

mapping_inline.cue
jpower432: @hbraswelrh Could you expand? Step is a string right now and making that a complex object would be a breaking change. I might be misunderstanding.

jpower432:
> ### 1. Naming: `origin` vs alternatives
> `origin` is concise, but in a git-centric open source ecosystem it strongly connotes "remote repository." When someone reads `origin: reference-id: github-api`, there is a moment of disambiguation. Have you considered `source` or `source-ref`? `source` reads more naturally for "where this evidence was collected from" and avoids the git overload. Not a blocker -- just worth a quick sanity check since renaming after stable is expensive.

I did think through `source` in the linked issue a bit. We also use it in `MappingDocument` so I was concerned about conflation there, but maybe it `source` actually aligns with the existing pattern and would reduce term sprawl. I would lean toward `source` and not `source-ref` so we can avoid kebob case where it is not needed for clarity.

jpower432:
> ### 3. Single vs multiple origins
> If a compliance finding is derived by correlating two sources (e.g., an API response and a config file), the singular `origin` field requires either splitting the finding into separate evidence entries or losing one of the source references. Has there been discussion about whether `origin` should be a list (`origins?: [#EvidenceMapping, ...]`)? The singular form is simpler and probably covers the majority of cases, but it would be good to confirm this is a deliberate choice rather than something that needs revisiting later.

This adds a mapping to an existing `Evidence` object. One `AssessmentLog` can reference multiple pieces of `Evidence`. Currently we only support representing the contents inline using `payload`. The `EvidenceMapping` alters the `Evidence` to allow the contents to be referenced from a remote source instead (as an additive option).

jpower432:
Both good callouts here.

`coordinate` is one of the items that I think it would be preferable to use a comment. We already have an `Evidence.type` that could likely be used to inform automated partial refetch workflows, but there are multiple gaps in the schema to achieve this that could be filled by the SDK. The primary purpose of this was to be a user hint that can support refetching. 

You are right `digest` could probably benefit from being constrained.

> ### 2. `coordinate` ambiguity
> The `coordinate` field is intentionally flexible, but consumers will receive values like `/orgs/foo/settings` (API path), `line 42` (file location), or `$.results[0].status` (JSONPath) with no way to distinguish the format programmatically. This works for human readers, but tools that need to render or cross-reference evidence locations would need to guess the format from context.
> 
> Two lightweight options that preserve simplicity:
> 
> * Add an optional `coordinate-type` field (`"api-path"`, `"json-pointer"`, `"file-location"`, etc.) so consumers can parse without guessing.
> * Document a convention in the comment (e.g., "URI-style coordinates are recommended") even if not schema-enforced.
> 
> This could also be deferred to a follow-up if you want to keep the initial surface area small and see how adopters actually use `coordinate` in practice.


>### 4. digest format
>The digest field has no format constraint. Since the test data uses sha256:a1b2c3d4e5f6..., a regex like =~"^[a-z0-9]+:>>[a-f0-9]+$" would enforce the algorithm:hex convention and align with OCI digest formats. This prevents >malformed values from passing validation silently. If there are use cases for non-standard digest formats, >documenting the expected format in the comment would help.

marcusburghardt (approved):
LGTM. Thanks for working on this @jpower432

mapping_inline.cue
hbraswelrh: @jpower432 this was a non-blocking comment and entirely from a "food for thought" perspective on whether the `step` field would be expanded to add more context to support the current `string` reference. I wasn't suggesting it to include in this PR.

eddie-knight (approved):

PR #454, opened 2026-08-07
fix: require audit results to reference declared criteria

eddie-knight:
@jpower432 I think this is correct, but please crazy check me here.


jpower432:
I did some local testing. This looks right to me as well. Did not affect codegen.

jpower432 (approved):

PR #459, opened 2026-08-24
fix(ai-agent): complete the ATR category mapping and repair its target reference

eeee2345:
@jpower432 — two defects in the artifact from #428, found while building the follow-on we discussed there. Split out as a fix so it does not sit behind the design conversation on the catalogs.

**The target reference does not resolve.** `target-reference.reference-id` and its entry in `metadata.mapping-references` both say `ATR-AI-AGENT-CAP`, but the capability catalog's `metadata.id` is `ATR-AI-AGENT-CAPABILITY-CATALOG`. `#MappingReference` says that when `url` is absent the id is the referenced artifact's `metadata.id`, so as merged this document points at nothing. Both occurrences fixed.

**The mapping is a category short.** ATR's `rules/` on main has ten detection categories; this maps nine. `model-security` was added upstream after #428 merged and never got an entry, so the artifact under-reports its own coverage without saying so.

Added, with the caveat stated rather than papered over: of that category's rules, only behaviour extraction relates to a capability this catalog models. The other two — poisoned fine-tuning data, and unsafe deserialization when a model artifact is loaded — are model supply chain, and there is no capability here for either. The rationale says that instead of reaching for an adjacent capability to make the row look complete. Proposing the missing capabilities is a separate change.

**Also drops "nine" from the description.** Not changed to "ten" — a count in prose goes stale the next time upstream adds a category, which is precisely how this one became wrong.

Verified: `go test -run TestSchemaValidation` passes, and every capability id the mapping references now resolves against the capability catalog.

Unrelated, but worth knowing since you will see it locally: `TestNoBreakingChanges` currently fails on `#EvaluationLog` and `#ControlEvaluation` against released v1.5.0. It fails on a clean `main` too, so it is not from this branch.


eeee2345:
@jpower432 — the red X on this PR is not from this branch, and the detail is probably worth having.

CI fails on `TestNoBreakingChanges/#EvaluationLog` and `TestNoBreakingChanges/#ControlEvaluation`, both reporting `field "assessment-logs" not present` when `main` is compared against the released `v1.5.0`. I reproduced it on a clean checkout of `main` at `9d36c25` with no changes in the tree, and it fails there the same way.

The timeline points at the release rather than at any branch:

| | |
| :--- | :--- |
| Last green CI | 2026-08-14 20:24Z, `fix/start-required` |
| `v1.5.0` published | 2026-08-14 20:57Z, 33 minutes later |
| CI run 2026-08-20 | `dependabot/github_actions/dependencies` — same two subtests fail |
| CI run 2026-08-24 | this branch — same two subtests fail |

Those two are the only CI runs since the release, and both failed. The dependabot one is the useful data point: it bumps GitHub Actions dependencies and cannot touch a CUE schema, so the compatibility gate has been failing for anything that runs it since `v1.5.0` shipped.

Reading it as a schema question rather than a test question: `#ControlEvaluation` gained `assessment-logs`, and the released `v1.5.0` does not have it, so the gate is doing its job — the incompatibility is real, it just predates this branch by ten days and no PR had run CI in between to surface it.

Nothing here needs anything from this PR; flagging it so the failure is not read as coming from these two lines of YAML. Happy to open a separate issue with the reproduction if that is more useful than a comment on an unrelated PR.

jpower432:
Thanks @eeee2345. Seeing the failure on other PRs as well. I will get a fix up to unblock this.

jpower432:
@eeee2345 A rebase should fix the failing CI

eeee2345:
@jpower432 Rebased onto main (now 0 behind, same single commit, DCO sign-off preserved). On the rebased branch locally: `make test` passes (schema validation suite complete) and `make lintinsights` is clean. Force-pushed; CI should re-run against current main. Sorry for the two-week gap on my side.

jpower432:
No worries @eeee2345! Thanks for the update.

jpower432 (approved):

PR #460, opened 2026-08-24
chore(compat): refactor compatibility test to use oasdiff

jpower432:
## Description

This PR swaps the breaking changes testing approach to use `oasdiff`. `oasdiff` is looking at the main types of breaking changes we need to capture - field adds, removals, adds/loosening existing constraints.

### Rationale
We have a CLI that already transforms CUE to OpenAPI. [`oasdiff`](https://github.com/oasdiff/oasdiff) is a tool that tests for OpenAPI diffs and breaking changes. It tests from the producer and consumer side and does not require as many workaround as `cue.Subsume`. It also has clearer error messages.

### Schema Changes Made

- [X] No schema changes
- [ ] Other

### Schema Change Details

```
```

## Testing

- [X] Manual testing performed

## Related Issues

Refs: #459 
Unblocks this and other in-flight PRs

## Reviewer Hints

 ---

## Self-review checklist


cmd/internal/cmd/breaking_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	cases := []struct {
+		name   string
+		rev    string
+		wantID string // "" means expect zero ERR changes
+	}{
+		{"identical", "testdata/base.yaml", ""},
+		{"removed_required", "testdata/rev_remove.yaml", "response-property-became-optional"},
+		{"added_required", "testdata/rev_add.yaml", "new-required-request-property"},
+		{"enum_narrowed", "testdata/rev_enum_narrow.yaml", "request-property-enum-value-removed"},
```
eddie-knight, line 29: 🤖 **Critical** — This case proves oasdiff can detect enum narrowing on a hand-written fixture, but the real pipeline can never produce that signal: `converter.go` lowers every disjunction to bare `type: string` and `SchemaInfo` has no enum field, so the generated `openapi.yaml` contains zero `enum:` keys. Verified empirically: narrowing `#ArtifactType` 13→3, `#Result` 6→2, or `#EntryType` 9→1 passes `breaking-check`, while the deleted `TestNoBreakingChanges` fails the identical tree. Since the PR's stated scope includes “adds/loosening existing constraints”, this is the gap that matters most for this spec — enum members are its core contract. The fix is local: emit `enum` for string-literal disjunctions in the converter and add an integration case that goes through the real generator; failing that, disclose the blind spot in CONTRIBUTING and track it.

Makefile
```diff
@@ -68,9 +68,23 @@ genopenapi:
...
+REPO ?= gemaraproj/gemara
+
 breaking-check:
-	@echo "  >  Running backward compatibility check ..."
-	@cd test && go test -v -run TestNoBreakingChanges ./...
+	@echo "  >  Checking for breaking schema changes ..."
+	@tag=$$(gh api "repos/$(REPO)/releases" --paginate \
+		--jq '.[] | select(.prerelease == false) | select(.tag_name | test("^v1\\.")) | select(any(.assets[]?; .name == "openapi.yaml")) | .tag_name' \
+		| sort -V | tail -n1); \
+	if [ -z "$$tag" ]; then \
```
eddie-knight, line 83: 🤖 **Major** — If `gh api` fails (auth lapse, rate limit, outage, missing binary), `$tag` is empty and this branch prints “No v1 baseline release… found; skipping” with exit 0 — the CI gate goes green with a message that misstates the cause. Reproduced with a stub `gh` exiting 1: `make breaking-check` returns 0. Capture `gh`'s exit status and fail on non-zero, skipping only when the query genuinely returns an empty list.

.github/workflows/ci.yml
```diff
@@ -25,3 +25,7 @@ jobs:
           go-version: stable
       - name: Run Schema Test
         run: make test
+      - name: Check breaking changes
+        run: make breaking-check
```
eddie-knight, line 29: 🤖 **Major** — Nothing runs the new tests: `make test` is `cd test && go test`, and no workflow or make target runs `go test` in `cmd/`, so the 259 lines of tests added under `cmd/internal/cmd/` never execute in CI. Add a step (or fold into `make test`) running `cd cmd && go test ./...`.

Makefile
```diff
@@ -68,9 +68,23 @@ genopenapi:
...
+		echo "  >  No v1 baseline release with an openapi.yaml asset found; skipping breaking-change check"; \
+		exit 0; \
+	fi; \
+	echo "  >  Comparing against v1 baseline $$tag"; \
+	tmp=$$(mktemp -d); \
+	trap 'rm -rf "$$tmp"' EXIT; \
+	gh release download "$$tag" --repo "$(REPO)" --pattern openapi.yaml --dir "$$tmp"; \
+	( cd cmd && go run . breaking-check --schema .. --base "$$tmp/openapi.yaml" )
```
eddie-knight: 🤖 **Major** — CONTRIBUTING's sanctioned escape hatch says to add the check ID to an allowlist file “and pass it to the check via `--allow`”, but this invocation never passes `--allow` and no allowlist path is wired anywhere in the repo — a contributor following the documented process would also have to patch this Makefile. Wire a conventional path (e.g. pass `--allow .oasdiff-allow` when the file exists) or correct the docs. Related scoping problem on `filterAllowed` in `breakingcheck.go` (separate comment).

cmd/internal/cmd/breaking.go
```diff
@@ -0,0 +1,72 @@
...
+func breakingChanges(base, rev *openapi3.T) ([]Change, error) {
+	d, sources, err := diff.GetWithOperationsSourcesMap(diff.NewConfig(),
+		&load.SpecInfo{Url: "base", Spec: base},
+		&load.SpecInfo{Url: "rev", Spec: rev})
+	if err != nil {
+		return nil, err
+	}
+	loc := checker.NewDefaultLocalizer()
+	var out []Change
+	for _, c := range checker.CheckBackwardCompatibility(checker.NewConfig(checker.GetAllChecks()), d, sources) {
```
eddie-knight, line 100: 🤖 **Major** — The gate now freezes `@status("experimental")` schemas as hard as stable ones: the deleted `compat_test.go` skipped every non-`@status("stable")` file (line 286 at the merge-base), but nothing here consumes the `x-status` the converter already emits per schema. Verified: flipping a field in an experimental-status schema fails the gate. Every experimental-artifact iteration now goes through the major-bump-or-allowlist ceremony that experimental status exists to avoid. oasdiff natively honors `x-stability-level` on operations — mapping `x-status: experimental` to `x-stability-level: alpha` on the synthesized operations in `loadWrapped` restores the old scoping. If freezing everything is intentional, state it in CONTRIBUTING.

cmd/internal/cmd/breakingcheck.go
```diff
@@ -0,0 +1,151 @@
...
+// breakingCheckOpts configures a breaking-change check run.
+type breakingCheckOpts struct {
+	schemaDir     string // CUE package directory used to generate the candidate
+	basePath      string // Baseline OpenAPI file to diff against (required)
+	candidatePath string // Local candidate OpenAPI file; empty means generate from schemaDir
+	allowPath     string // Allowlist file of oasdiff check IDs; empty means no allowlist
+}
+
+// filterAllowed drops changes whose ID appears in allowedIDs.
+func filterAllowed(changes []Change, allowedIDs map[string]bool) []Change {
```
eddie-knight: 🤖 **Major** — Keying the allowlist on check ID alone turns a narrowly scoped exception into a global one: a single entry (say `request-property-enum-value-removed`) disables that entire check class for every schema, stable included, for as long as the entry lives — the opposite of CONTRIBUTING's “narrowly scoped, well-understood exceptions”. `Change` carries no path/schema, so no finer key exists today. Carry `c.GetPath()` through `Change` and key entries as `<id> <path>` (optionally still accepting bare `<id>`).

cmd/internal/cmd/breaking.go
```diff
@@ -0,0 +1,72 @@
...
+// breaking).
+func loadWrapped(path string) (*openapi3.T, error) {
+	doc, err := openapi3.NewLoader().LoadFromFile(path)
+	if err != nil {
+		return nil, err
+	}
+	if doc.Paths == nil {
+		doc.Paths = openapi3.NewPaths()
+	}
+	for name := range doc.Components.Schemas {
```
eddie-knight, line 40: 🤖 **Minor** — `doc.Components` is nil when the file has no `components:` section, and this line panics on it (reproduced with a minimal spec). Both inputs are generated by this repo's own tooling today, but the baseline is a downloaded release asset — cheap to guard.

```suggestion
	if doc.Components == nil {
		doc.Components = &openapi3.Components{}
	}
	for name := range doc.Components.Schemas {
```

cmd/internal/cmd/breaking.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/oasdiff/oasdiff/checker"
+	"github.com/oasdiff/oasdiff/diff"
+	"github.com/oasdiff/oasdiff/load"
+)
+
+// Change is an ERR-level backward-incompatible change reported by oasdiff.
+type Change struct {
+	ID   string
```
eddie-knight: 🤖 **Minor** — Dropping oasdiff's path context makes multi-hit output unattributable: a run against a change touching several schemas prints the same `ERR[new-required-request-property]: added the new required request property metadata/newthing` line repeatedly with no schema name. Since “clearer error messages” is this PR's stated rationale, carry `c.GetPath()` through `Change` and include it in the printout — this also unblocks the per-schema allowlist keying suggested in `breakingcheck.go`.

.github/workflows/ci.yml
```diff
@@ -25,3 +25,7 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
eddie-knight, line 31: 🤖 **Minor** — The token is handed to a step whose recipe ends by compiling and running PR-authored Go (`go run .` in `cmd/`), but everything it fetches is public. Swapping the two `gh` calls for `curl -fsSL` against the public REST/download URLs lets you drop this env line entirely (plain `gh` refuses to run tokenless in Actions), shrinking what PR code can reach to nothing. Fork PRs already get a read-only token, so this is hardening rather than a hole; the trade-off is unauthenticated calls share the runner-IP rate limit.

Makefile
```diff
@@ -68,9 +68,23 @@ genopenapi:
...
 #
+# Repo whose releases hold the v1 baseline OpenAPI asset.
+REPO ?= gemaraproj/gemara
+
 breaking-check:
-	@echo "  >  Running backward compatibility check ..."
-	@cd test && go test -v -run TestNoBreakingChanges ./...
+	@echo "  >  Checking for breaking schema changes ..."
+	@tag=$$(gh api "repos/$(REPO)/releases" --paginate \
+		--jq '.[] | select(.prerelease == false) | select(.tag_name | test("^v1\\.")) | select(any(.assets[]?; .name == "openapi.yaml")) | .tag_name' \
```
eddie-knight: 🤖 **Minor** — Hardcoding `select(.prerelease == false)` silently drops the `GEMARA_COMPAT_PRERELEASE=true` knob the deleted test supported (`compat_test.go:314` at the merge-base) and that existing docs reference. Either thread an equivalent through (skip this select when the variable is set) or note the removal so stale references get cleaned up.

cmd/internal/cmd/breakingcheck.go
```diff
@@ -0,0 +1,151 @@
...
+	opts := breakingCheckOpts{}
+	cmd := &cobra.Command{
+		Use:   "breaking-check",
+		Short: "Detect backward-incompatible schema changes against the v1 baseline",
+		Long: `Generate an OpenAPI projection of the current CUE schema and diff it against
+a baseline OpenAPI file using oasdiff. The baseline is supplied via --base
+(fetched from the latest v1 release by the Makefile/CI). Backward-incompatible
+(ERR-level) changes fail the check unless their oasdiff check ID is listed in the
+allowlist.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
eddie-knight, line 139: 🤖 **Nit** — On a detected breaking change, cobra prints the full usage block after “Error: breaking changes detected”, burying the findings it just printed.

```suggestion
		SilenceUsage: true,
		RunE: func(cmd *cobra.Command, args []string) error {
```

.github/workflows/release.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: '1.25'
+          go-version: stable
```
eddie-knight: 🤖 **Nit** — This trades a pinned release toolchain for a floating one. It matches ci.yml, but the release job is exactly where a surprise toolchain bump hurts; `'1.26'` satisfies the new deps and keeps the pin.

```suggestion
          go-version: '1.26'
```

eddie-knight (commented):
I like the idea here! But there are some issues to hammer out in the implementation.

---

🤖 Multi-model review coordinated by Fable — three independent reviewers (Haiku, Sonnet, Opus tiers) over the full worktree at 400c9a7; every claim was re-verified against source before posting, empirically where noted.

### Focus questions

**1. Deep-check.** The implementation works for what it covers — verified end-to-end against the real v1.5.0 asset (field add/remove across nested `$ref` chains, required flips, type changes, schema removal, both directions; the `loadWrapped` synthesis is necessary since oasdiff's components-level check is INFO-only). Three verified problems: the inline Critical (the gate cannot see enum narrowing — removing `#ArtifactType` members passes CI while the deleted test failed it) and two inline Majors (a `gh` failure silently passes the gate; the new `cmd/` tests never run in CI).

**2. Better dependency option?** No. oasdiff is Apache-2.0, actively maintained (pushed the day of this review), and the live successor to the Tufin lineage; the alternatives (pb33f/openapi-changes, the oasdiff Action, keeping `cue.Subsume`) each fit worse. The real ceiling is not the differ but `converter.go`: no OpenAPI tool can see constraints the projection never emits (enums, numeric types, most patterns), so the Critical's fix belongs in the converter.

**3. Above reproach?** On supply chain, yes: licenses Apache-2.0/MIT-compatible, deps version-pinned with go.sum checksums, actions SHA-pinned, `pull_request` trigger with read-only fork token (one hardening note inline). The gap is the gate's honesty: CONTRIBUTING declares the schemas "a public contract" without disclosing the enum blind spot, and the gate now freezes `@status("experimental")` schemas the old test deliberately exempted. Closing that doc/behavior gap is the main pre-merge ask.

### Findings
1 Critical, 5 Major, 4 Minor, 2 Nit — all inline. Minimum bar: the Critical plus the two Majors on silent `gh` failure and unwired tests; the allowlist pair matters before the escape hatch is first used.

### Checked and NOT problems (recorded so they don't get re-raised)
- Verified fine: CI trust boundary, `test/go.mod` cleanup, untouched schema tests/fixtures, baseline-selection jq, same-tag CUE+OpenAPI publishing, removal of the old subsume workarounds.
- Refuted during verification: "oasdiff has no pattern rules at all" (it has them for request parameters, just not body schemas); "the coverage regression is documented in CONTRIBUTING" (the section never discloses it); the nil-`Components` panic was downgraded from a top-severity claim after reproduction — both inputs are self-generated in every real invocation.

.github/workflows/release.yml
jpower432: Accepted this, but we actually might just want to point at the `go.mod`

eddie-knight: Yeah, that would make sense

cmd/internal/cmd/breaking_test.go
jpower432: Is this being handled in #461?

jpower432: Just going to cherry-pick those commits so we can unblock the `Critical` on this PR.

eddie-knight: Let me know if that ends up resolving the need for #461. I haven't yet felt like that PR is complete, so happy to close it if we get the fixes in here.

Makefile
jpower432: Will removed the references. `GEMARA_COMPAT_PRERELEASE` was a pre-v1 knob.

.github/workflows/ci.yml
jpower432: Left this in place because `gh` is used to grab the release information too. That and the possibility of rate limiting with unauthenticated calls seems like enough to keep in. We could always add this to a new job and scope the token down to `contents: read`.

cmd/internal/cmd/breaking.go
jpower432: Fixed in https://github.com/gemaraproj/gemara/pull/460/changes/71a81e761602f87826cdedc30c4e21b074bd1d2b

cmd/internal/cmd/breaking.go
jpower432: Fixed in https://github.com/gemaraproj/gemara/pull/460/changes/71a81e761602f87826cdedc30c4e21b074bd1d2b

.github/workflows/ci.yml
jpower432: Added in cherry-pick from #461

cmd/internal/cmd/breakingcheck.go
jpower432: Good catch on this one. I added this and just dropped the bare `id`. I couldn't think of a case where we would need a global exclude except for when we used it for `Datetime` in CUE, but that was one of the workarounds.

Makefile
jpower432: Wired with this up and used in this PR because of the what we have for the baseline on release `v1.5.0`. Added temporary exceptions over modifying the release asset.

jpower432:
@eddie-knight Ready for re-review, but I am going to test the converter changes with the `website` automation to make sure it would break the generation of the schema pages.

**EDITED** Verified and the hidden defs work fixed the broken link between Control Evaluations and Assessment Logs on the website schema docs.

eddie-knight (approved):

PR #461, opened 2026-08-25
fix(cmd): emit enum values and filter hidden definitions in cue2openapi

eddie-knight:
## Description 🤖 

- Disjunctions of string literals now emit OpenAPI enum values (handles nested chains and *"default" markers).
- Hidden definitions (#_Name) are omitted from output; refs to them are rewritten to the visible definition they constrain.
- Hidden struct fields (_name) no longer leak as properties.
- make test now also runs the cmd/ test suite in CI.

### Downstream impact

Although no CUE schemas change, the generated `generated/openapi.yaml` (published as a release asset and consumed by `@gemara/react`'s type generation) changes shape: 17 schemas gain `enum` (generated TS types become literal unions instead of `string`), two `$ref`s move from `_MappingStrict`/`_AssessmentLogStrict` to `Mapping`/`AssessmentLog`, and two hidden schemas plus two hidden fields disappear from the output.

### Schema Changes Made

- [x] No schema changes
- [ ] Audit Log schema (`auditlog.cue`) changes
- [ ] Capability Catalog schema (`capabilitycatalog.cue`) changes
- [ ] Control Catalog schema (`controlcatalog.cue`) changes
- [ ] Enforcement Log schema (`enforcementlog.cue`) changes
- [ ] Evaluation Log schema (`evaluationlog.cue`) changes
- [ ] Guidance Catalog schema (`guidancecatalog.cue`) changes
- [ ] Mapping Document schema (`mappingdocument.cue`) changes
- [ ] Policy Document schema (`policy.cue`) changes
- [ ] Principle Catalog schema (`principlecatalog.cue`) changes
- [ ] Risk Catalog schema (`riskcatalog.cue`) changes
- [ ] Threat Catalog schema (`threatcatalog.cue`) changes
- [ ] Vector Catalog schema (`vectorcatalog.cue`) changes
- [ ] Other

### Schema Change Details

```
<!-- If applicable, provide a brief summary or example of schema changes -->
```

## Testing

- [ ] Unit tests added/updated
- [ ] Manual testing performed
- [ ] Test data updated (if applicable)

## Reviewer Hints

 ---

## Self-review checklist

- [ ] This PR has content that was created with AI assistance. 
   - [ ]  I have read and followed the [Generative AI Contribution Policy](https://www.linuxfoundation.org/legal/generative-ai).
- [ ] I have the experience and knowledge necessary to answer maintainer questions about the content of this PR, without using AI.

PR #478, opened 2026-09-05
fix(cue2openapi): generate OpenAPI with the upstream CUE encoder

### LINKED ISSUE #466: cue2openapi: generated Datetime pattern rejects valid RFC 3339 timestamps
---

> 🦾 Written with LLM assistance [claude-opus-5]<br/>
> 💪 Reviewed by a human before submission

---

## Version of Gemara

`v0.17.0-dev` (commit `ec2e3d055fe6f4e059ac43273379d3666bd7ec94`), reproduced with `cue v0.17.1`.

## Describe the bug

`cmd/cue2openapi` emits a `Datetime` schema whose `pattern` is date-only, while the CUE source is a full RFC 3339 date-time. As a result the generated `openapi.yaml` rejects the timestamps in the repository's own test data.

Source, `metadata.cue:13`:

```cue
#Datetime: time.Format("2006-01-02T15:04:05Z07:00") @go(Datetime,format="date-time")
```

Generated output:

```yaml
Datetime:
  type: string
  pattern: "^\\d{4}-\\d{2}-\\d{2}$"
  format: date
  x-status: stable
```

The layout specifies a date, a time, and a numeric offset. The generated pattern allows only the date.

The cause looks like `convertTypeAlias` in `cmd/internal/cmd/converter.go:612`. Its `*ast.CallExpr` branch matches any `time.Format(...)` call and returns a fixed date-only schema without reading `x.Args`, so the layout argument isn't consulted.

<details>
<summary>Relevant branch, <code>converter.go:640-649</code></summary>

```go
if ident, ok := sel.X.(*ast.Ident); ok && ident.Name == "time" {
    if selLabel, ok := sel.Sel.(*ast.Ident); ok && selLabel.Name == "Format" {
        return &SchemaInfo{
            Type:        "string",
            Description: description,
            Format:      "date",
            Pattern:     "^\\d{4}-\\d{2}-\\d{2}$",
        }
    }
}
```

Because the layout is never inspected, any `time.Format` layout in any schema would be converted the same way.

</details>

## To Reproduce

```console
$ git clone https://github.com/gemaraproj/gemara && cd gemara
$ git rev-parse HEAD
ec2e3d055fe6f4e059ac43273379d3666bd7ec94

$ make genopenapi
$ grep -A4 '^    Datetime:' generated/openapi.yaml
    Datetime:
      type: string
      description: Datetime represents an ISO 8601 formatted datetime string
      pattern: "^\\d{4}-\\d{2}-\\d{2}$"
      format: date
```

Then check that pattern against the committed fixture. CUE accepts the file; the generated pattern rejects all three of its timestamps.

<details>
<summary>Full transcript</summary>

```console
$ grep -o 'collected-at: "[^"]*"' test/test-data/good-audit-log.yaml
collected-at: "2025-08-22T16:02:00Z"
collected-at: "2025-08-22T16:05:00Z"
collected-at: "2026-02-10T15:05:00Z"

$ python3 -c "
import re
p = re.compile(r'^\d{4}-\d{2}-\d{2}\$')
for v in ['2025-08-22T16:02:00Z','2025-08-22T16:05:00Z','2026-02-10T15:05:00Z']:
    print(v, '->', 'MATCH' if p.match(v) else 'REJECTED')
"
2025-08-22T16:02:00Z -> REJECTED
2025-08-22T16:05:00Z -> REJECTED
2026-02-10T15:05:00Z -> REJECTED

$ cue vet -d '#AuditLog' ./*.cue test/test-data/good-audit-log.yaml
$ echo $?
0
```

</details>

## Expected behavior

The generated OpenAPI should describe the same value space as the CUE. For `time.Format("2006-01-02T15:04:05Z07:00")` that would be `format: date-time`, matching the `@go(Datetime,format="date-time")` attribute already on the field, with either an RFC 3339 pattern or no `pattern` at all.

More generally, it'd help if `convertTypeAlias` read the layout from `x.Args` and mapped it, and surfaced an error for a layout it can't map rather than falling back to a narrower pattern.

## Additional context

Two consumers sit downstream of this file, so the effect reaches past the reference docs.

<details>
<summary>Downstream consumers</summary>

`.github/workflows/release.yml:69` publishes it as a release asset:

```yaml
run: gh release upload "$TAG" generated/openapi.yaml --clobber
```

The `notify-website` job just below it describes one consumer:

```
# The website repo regenerates its schema reference pages from the
# released spec, so tell it a new release exists.
```

And the `Makefile` notes another:

```
# downstream type generation (e.g. gemara-react) consumes generated/openapi.yaml.
```

So a validator or type generated from the release asset would reject conformant documents, and the same pattern renders at <https://gemara.openssf.org/schema/base#datetime>, which is where I first noticed it.

</details>

`#Datetime` is `@status("stable")`, so this sits on the stable surface.

I filed a second, separate fidelity issue for `#Evidence.payload` in the same file. It's a different code path but might be worth fixing in the same pass. PR #461 is already working nearby in `cue2openapi`, so there may be overlap there too.

### LINKED ISSUE #468: cue2openapi: Evidence.payload emitted as string though the CUE is the top type
---

> 🦾 Written with LLM assistance [claude-opus-5]<br/>
> 💪 Reviewed by a human before submission

---

## Version of Gemara

`v0.17.0-dev` (commit `ec2e3d055fe6f4e059ac43273379d3666bd7ec94`).

## Describe the bug

`#Evidence.payload` accepts any value. From `auditlog.cue:92`:

```cue
payload?: _ @go(Payload,type=any)
```

The generated OpenAPI declares it a string:

```yaml
Evidence:
  properties:
    payload:
      type: string
      description: payload is the raw evidence data collected inline
```

Structured inline evidence is legal, and it's the form used throughout ADR-0023 and ADR-0024, for example `payload: {vulnerability_reporting_enabled: true}`. Reading the published schema, an implementer would reasonably conclude that inline evidence has to be serialized into a string, which produces documents that don't match the ADRs' own examples.

The cause looks like the fallback branches in `cmd/internal/cmd/converter.go`, which return `&SchemaInfo{Type: "string"}` when an expression isn't recognized. `_` isn't handled anywhere, so it reaches one of those.

<details>
<summary>String fallbacks in <code>converter.go</code></summary>

```
367:  schemaInfo := &SchemaInfo{Type: "string", Description: description}   // "// Fallback: create a basic string schema"
528:  return &SchemaInfo{Type: "string", Description: description}          // convertExprToSchema default
535:  return &SchemaInfo{Type: "string", Description: description}
543:  return &SchemaInfo{Type: "string", Description: description}          // convertIdentToSchema default
576:  return &SchemaInfo{Type: "string", Description: description}
579:  return &SchemaInfo{Type: "string", Description: description}          // convertBinaryExprToSchema default
653:  return &SchemaInfo{Type: "string", Description: description}          // convertTypeAlias default
```

</details>

## To Reproduce

```console
$ git clone https://github.com/gemaraproj/gemara && cd gemara
$ git rev-parse HEAD
ec2e3d055fe6f4e059ac43273379d3666bd7ec94

$ make genopenapi
$ sed -n '/^    Evidence:/,/^      required:/p' generated/openapi.yaml
```

Observed, against `auditlog.cue:92` which reads `payload?: _`:

```yaml
Evidence:
  type: object
  properties:
    payload:
      type: string
      description: payload is the raw evidence data collected inline
```

## Expected behavior

`payload` emitted as an unconstrained schema, so objects, arrays, numbers, booleans, and strings all validate. In OpenAPI 3.0 that's `payload: {}` with a description; 3.1 allows `true`. Either would match the `@go(Payload,type=any)` attribute already on the field.

It'd also be worth considering whether the fallback branches should emit `type: string` at all. An unconstrained schema, or a hard error, would surface conversion gaps instead of narrowing the contract quietly. The `Datetime` issue I filed separately is the same class of problem with larger consequences.

## Additional context

This is what renders at <https://gemara.openssf.org/schema/auditlog#evidence>, which lists `payload` as a string.

One related observation, possibly for `gemaraproj/website` rather than here: on that page the `source` field's `EvidenceMapping` type isn't hyperlinked and has no reference page, even though `EvidenceMapping` is present in `generated/openapi.yaml` under `components.schemas`. That looks like a navigation gap rather than a generator bug. Mentioning it only because it makes `source` harder to discover, and `source` is the field you need for evidence too large to inline.

PR #461 is working adjacent ground in `cue2openapi`.

### LINKED ISSUE #473: cue2openapi: aliased and embedded fields are missing from the generated schema
---

> 🦾 Written with LLM assistance [claude-opus-5]<br/>
> 💪 Reviewed by a human before submission

---

## Version of Gemara

`v0.17.0-dev` (commit `ec2e3d055fe6f4e059ac43273379d3666bd7ec94`). Also reproduced against the published `openapi.yaml` release asset for `v1.5.0`.

## Describe the bug

Two code paths in `cmd/cue2openapi` skip fields while walking the CUE AST, so the generated `openapi.yaml` describes fewer fields than the schema it was generated from. Some of the missing fields are required.

The result is that a document authored against the published spec can omit fields that `cue vet` then rejects, and a type generated from it has no way to express them at all.

### 1. Fields with a label alias are dropped

`getFieldName` handles `*ast.Ident` and `*ast.BasicLit` labels. A label written as `X="field-name"` parses as an `*ast.Alias`, falls through the switch, and returns the empty string. The caller at `converter.go:468` gates on `fieldName != ""`, so the field is skipped without a diagnostic.

<details>
<summary><code>converter.go:656-666</code> and its caller</summary>

```go
func getFieldName(field *ast.Field) string {
	switch label := field.Label.(type) {
	case *ast.Ident:
		return label.Name
	case *ast.BasicLit:
		if label.Kind == token.STRING {
			return strings.Trim(label.Value, "\"")
		}
	}
	return ""
}
```

```go
fieldName := getFieldName(x)
// ...
if fieldName != "" && !hidden {
	schema.Properties[fieldName] = fieldSchema
	if x.Optional == token.NoPos {
		schema.Required = append(schema.Required, fieldName)
	}
}
```

</details>

The schema has exactly two aliased labels, both on `#Metadata` at `metadata.cue:51` and `metadata.cue:54`:

```cue
MR="mapping-references"?: [#MappingReference, ...#MappingReference] @go(MappingReferences) @yaml("mapping-references,omitempty")
AG="applicability-groups"?: [#Group, ...#Group] @go(ApplicabilityGroups) @yaml("applicability-groups,omitempty")
```

Both are absent from the published spec. `mapping-references` is what every `reference-id` in the model resolves into, across `#ArtifactMapping`, `#EntryMapping`, `#MultiEntryMapping`, and `#EvidenceMapping`. `reference-id` appears 14 times across four schema files, and the field they all point at isn't in the published schema.

### 2. Embedded definitions are not flattened

`convertStructToSchema` iterates `st.Elts` and switches on `*ast.Field` and `*ast.CommentGroup` at `converter.go:457`. An embedded definition is an `*ast.EmbedDecl`, which matches neither, so its fields never reach the generated schema. `converter.go:395` is the only place `*ast.EmbedDecl` appears, inside the `findBaseIdent` helper, and that path doesn't contribute properties.

The schema embeds in 15 places. `#Catalog` is embedded by all seven catalogs, `#Log` by the three log types, `#Entity` by `#Actor` and `#Resource`, `#ArtifactMapping` by `#MultiEntryMapping` and `#TypedMapping`, and `#Group` by `#RiskCategory`.

`#Catalog` contributes `title`, `metadata`, `groups`, `extends`, and `imports`. `#Log` contributes `metadata` and `target`. Of those, `title` and `target` are required.

| Published schema    | CUE also defines                                      | Required in CUE      |
| ------------------- | ----------------------------------------------------- | -------------------- |
| `ControlCatalog`    | `title`, `groups`, `extends`, `imports`               | `title`              |
| `ThreatCatalog`     | `title`, `groups`, `extends`, `imports`               | `title`              |
| `AuditLog`          | `target`                                              | `target`             |
| `EvaluationLog`     | `target`                                              | `target`             |
| `Actor`             | `id`, `name`, `type`, `version`, `description`, `uri` | `id`, `name`, `type` |
| `Resource`          | `id`, `name`, `type`, `version`, `description`, `uri` | `id`, `name`, `type` |
| `MultiEntryMapping` | `reference-id`, `remarks`                             | `reference-id`       |
| `RiskCategory`      | `id`, `title`, `description`                          | all three            |

Several of these compound. `#Log.target` is a `#Resource`, so the field is missing from the log types, and the schema it would have pointed at is itself missing the six fields it embeds from `#Entity`. `MultiEntryMapping` losing `reference-id` lands on the same mechanism as the missing `metadata.mapping-references` above: from the published spec alone, neither end of a cross-artifact reference can be expressed.

`metadata` survives on the artifact types only because each one re-declares it in its own body, for example `metadata: type: "AuditLog"` at `auditlog.cue:14`. It isn't arriving through the embedding.

## To Reproduce

Against the published release asset, so no build is needed:

```console
$ curl -sL https://github.com/gemaraproj/gemara/releases/download/v1.5.0/openapi.yaml -o openapi.yaml

$ python3 -c "
import yaml
s = yaml.safe_load(open('openapi.yaml'))['components']['schemas']
print('Metadata      ', sorted(s['Metadata']['properties']))
print('ControlCatalog', sorted(s['ControlCatalog']['properties']))
print('AuditLog      ', sorted(s['AuditLog']['properties']))
print('Actor         ', sorted(s['Actor']['properties']))
"
Metadata       ['author', 'date', 'description', 'draft', 'gemara-version', 'id', 'lexicon', 'type', 'version']
ControlCatalog ['controls', 'metadata']
AuditLog       ['criteria', 'metadata', 'owner', 'results', 'summary']
Actor          ['contact']
```

`make genopenapi` at `ec2e3d0` produces the same output.

<details>
<summary>Isolating the alias case</summary>

Two fields of identical type in the same struct, differing only by the alias. Add both to `#Metadata` in `metadata.cue` and run `make genopenapi`:

```cue
"probe-no-alias"?:      [#Group, ...#Group]
PB="probe-with-alias"?: [#Group, ...#Group]
```

```
probe-no-alias           PRESENT
probe-with-alias         ABSENT
```

(`cue vet` reports `unreferenced alias or let clause PB` for the probe, since nothing reads `PB`. That doesn't affect `make genopenapi`, which is what's being measured here.)

Unwrapping the alias in `getFieldName` restores both aliased `#Metadata` fields:

```go
// X="field-name": the alias wraps the real label; unwrap before naming.
var expr ast.Node = field.Label
if alias, ok := field.Label.(*ast.Alias); ok {
	expr = alias.Expr
}
switch label := expr.(type) {
// ... unchanged
```

```
mapping-references       PRESENT
applicability-groups     PRESENT
```

`ast.Alias.Expr` is an `ast.Expr` rather than an `ast.Label`, so it needs a separate switch rather than reassignment into `field.Label`'s type.

</details>

## Expected behavior

The generated OpenAPI should describe the same fields as the CUE.

For aliases, `getFieldName` should unwrap `*ast.Alias` and use the underlying label. The alias is a reference mechanism for use elsewhere in the struct, not part of the field's name.

For embeddings, `convertStructToSchema` should resolve `*ast.EmbedDecl` and merge the embedded definition's properties and required list into the enclosing schema. Emitting an `allOf` against the embedded schema would also work and would keep the relationship visible.

Either way, a construct the converter doesn't understand would be better surfaced as an error than skipped silently. Both of these produce a schema that looks complete and isn't.

## Additional context

`#Metadata` and `#Catalog` are both `@status("stable")`, so this sits on the stable surface. `v1.5.0` is a `v1` release, which is the baseline `make breaking-check` compares against.

Two consumers sit downstream of this file. `.github/workflows/release.yml:69` publishes it as a release asset, the `notify-website` job tells the website repo to regenerate its schema reference pages from it, and the `Makefile` notes that "downstream type generation (e.g. gemara-react) consumes generated/openapi.yaml".

I've filed a companion issue covering constraints that survive generation but come out wrong, which is a different failure of the same file. Two earlier issues cover `#Datetime` and `#Evidence.payload` in the same converter. All four look like separate code paths rather than one root cause, though they may be worth fixing in one pass. PR #461 is already working nearby.

### LINKED ISSUE #474: cue2openapi: regex constraints are dropped or double-escaped in the generated schema
---

> 🦾 Written with LLM assistance [claude-opus-5]<br/>
> 💪 Reviewed by a human before submission

---

## Version of Gemara

`v0.17.0-dev` (commit `ec2e3d055fe6f4e059ac43273379d3666bd7ec94`). Also reproduced against the published `openapi.yaml` release asset for `v1.5.0`.

## Describe the bug

The schema constrains five values with a regex. In the generated `openapi.yaml`, four lose the constraint entirely and the fifth comes out escaped one level too deep, which changes what it matches.

The published spec is therefore both more permissive than the CUE in four places and, in the remaining one, rejects every value the CUE accepts.

### 1. Named-definition regexes are escaped twice

`#Email` at `entities.cue:68`:

```cue
#Email: =~"^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}$"
```

In CUE, `"\\."` is the two-character regex `\.`, an escaped literal dot. Published output, at `openapi.yaml:628`:

```yaml
Email:
  type: string
  description: Email represents a validated email address pattern
  pattern: "^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\\\.[A-Za-z]{2,}$"
  x-status: stable
```

That's a double-quoted YAML scalar carrying four backslashes, which unescape to two. So a consumer loads the pattern as `\\.` where the CUE means `\.`. As a regex `\\` matches one literal backslash and `.` then matches any character, so the address has to contain a backslash to validate. No conformant email does.

The cause looks like the converter trimming the quote characters off the CUE string literal instead of unquoting it, at `converter.go:310`, so the escape sequence survives into the emitted document.

<details>
<summary>Following the backslashes through</summary>

| Stage                                                      | Value                                     |
| ---------------------------------------------------------- | ----------------------------------------- |
| `entities.cue:68`, raw source text                         | `\\.`                                     |
| What CUE means by that literal                             | `\.`, an escaped literal dot              |
| `ast.BasicLit.Value` after `strings.Trim(lit.Value, "\"")` | `\\.`, unchanged                          |
| YAML-encoded at `openapi.yaml:628`                         | `\\\\.`                                   |
| What a YAML parser loads                                   | `\\.`                                     |
| What that regex matches                                    | one literal backslash, then any character |

`BasicLit.Value` is the raw source text with escape sequences intact, so trimming the quote characters leaves `\\.` where unquoting would give `\.`. There's no `strconv.Unquote` or `literal.Unquote` call anywhere in `converter.go`. The same shape appears at `:329`, `:331`, `:440`, `:553`, `:562`, `:620`, `:629`, and at `:662` in `getFieldName`, so a fix probably wants a shared helper.

</details>

### 2. Field-level regexes are dropped

A regex written directly on a field, rather than as a named definition, produces a bare `type: string` with no `pattern`. Four fields are affected:

| Field                     | Source                  | Constraint lost                                   |
| ------------------------- | ----------------------- | ------------------------------------------------- |
| `#Entity.uri`             | `entities.cue:27`       | `^https?://[^\s]+$`                               |
| `#LexiconReference.url`   | `lexicon.cue:48`        | `^(https?\|file)://[^\s]+$`                       |
| `#MappingReference.url`   | `mapping_inline.cue:24` | `^(https?\|file)://[^\s]+$`                       |
| `#EvidenceMapping.digest` | `mapping_inline.cue:74` | `^[a-z0-9]+(?:[+._-][a-z0-9]+)*:[a-zA-Z0-9=_-]+$` |

`digest` is the one I'd weight highest. It's the integrity anchor for referenced evidence, and the `algorithm:encoded` shape is the thing a consumer parses. From the published spec it's an unconstrained string.

The cause here looks separate from the escaping above: the field path doesn't recognize the AST node a field-level `=~` parses into, and falls back to a bare string.

<details>
<summary>Why the field-level path drops the pattern</summary>

CUE parses `=~"pattern"` as an `*ast.UnaryExpr`, which `convertExprToSchema` treats as an optional marker and recurses through at `converter.go:523-525`. That lands on the `*ast.BasicLit` holding the pattern, which the switch doesn't handle, so it falls to the `type: string` default at `:528` and the pattern is discarded.

`convertBinaryExprToSchema` at `:550-568` does handle `=~` and would emit the pattern correctly, but a field-level regex never reaches it. `parseDefinitionField` already accounts for this at `:304-305`, with the comment "CUE parses `=~"pattern"` as UnaryExpr with MAT op", which is why the named-definition path works and this one doesn't.

</details>

## To Reproduce

Against the published release asset:

```console
$ curl -sL https://github.com/gemaraproj/gemara/releases/download/v1.5.0/openapi.yaml -o openapi.yaml

$ python3 -c "
import yaml, re
s = yaml.safe_load(open('openapi.yaml'))['components']['schemas']
p = s['Email']['pattern']
print('published pattern:', p)
rx = re.compile(p)
for v in ['maintainer@example.com', 'a.b+c@sub.example.co.uk']:
    print(' ', v, '->', 'MATCH' if rx.match(v) else 'REJECTED')
"
published pattern: ^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}$
  maintainer@example.com -> REJECTED
  a.b+c@sub.example.co.uk -> REJECTED
```

The same addresses match once the pattern carries the single backslash the CUE source means.

```console
$ python3 -c "
import yaml
s = yaml.safe_load(open('openapi.yaml'))['components']['schemas']
for t, f in [('Entity','uri'), ('LexiconReference','url'),
             ('MappingReference','url'), ('EvidenceMapping','digest')]:
    print(f'{t}.{f}:', s[t]['properties'][f].get('pattern', 'no pattern'))
"
Entity.uri: no pattern
LexiconReference.url: no pattern
MappingReference.url: no pattern
EvidenceMapping.digest: no pattern
```

The schema writes five `=~` regexes, at `entities.cue:27` and `:68`, `lexicon.cue:48`, and `mapping_inline.cue:24` and `:74`. Exactly one of the five reaches the published document, and it arrives escaped wrong. The only other `pattern` key anywhere in the document is `Datetime`, which the converter synthesizes rather than reading from a source regex, and which is the subject of a separate issue.

`make genopenapi` at `ec2e3d0` produces the same output.

## Expected behavior

A regex should reach the generated schema unchanged, whether it's written as a named definition or directly on a field.

For the escaping, unquoting the CUE string literal rather than trimming its quote characters would fix it, and would apply to every other place the converter reads a `BasicLit` value.

For the dropped constraints, a field whose value is a `=~` expression should emit `type: string` with the corresponding `pattern`, matching what the named-definition path already attempts.

Where a construct can't be represented, an error would be better than silently emitting a weaker schema. The four dropped constraints look identical to a correctly generated unconstrained string.

## Additional context

`#Email`, `#Entity`, `#MappingReference`, and `#EvidenceMapping` are all `@status("stable")`.

This bears on a separate issue I filed about the URL scheme allowlist on `#MappingReference.url`. That issue argues the allowlist is too narrow, since it rejects `oci://` from ADR-0023's own example. Worth knowing alongside it: the allowlist isn't enforced by the published spec at all, so the two artifacts disagree about what a valid URL is. Whichever way the allowlist question is settled, the generated schema should reflect the answer.

Two consumers sit downstream of this file. `.github/workflows/release.yml:69` publishes it as a release asset, the `notify-website` job tells the website repo to regenerate its schema reference pages from it, and the `Makefile` notes that "downstream type generation (e.g. gemara-react) consumes generated/openapi.yaml". A validator built from the release asset would reject every `#Email` and would accept a malformed `digest`.

I've filed a companion issue covering fields that don't reach the generated schema at all, which is a different failure of the same file. Two earlier issues cover `#Datetime` and `#Evidence.payload` in the same converter. PR #461 is already working nearby.

jpower432:
## Description

This PR rewrite the `cue2openapi` command to CUE's upstream converter logic. This requires pre and post processing due to some limitations upstream (noted throughout the comments). Changes will require changes to website schema page generation.

This includes:
- A golden openAPI file
- Regressions tests for filed issues against the current logic

### Next Steps

File issues upstream to address workarounds

### Schema Changes Made

- [X] No schema changes
- [ ] Other

### Schema Change Details

```
```

## Testing

- [x] Unit tests added/updated
- [x] Test data updated (if applicable)

## Related Issues

Closes #466 
Closes #468 
Closes #474 
Closes #473 

Supersedes #461 

## Reviewer Hints

 ---

## Self-review checklist

- [x] This PR has content that was created with AI assistance. 
   - [x]  I have read and followed the [Generative AI Contribution Policy](https://www.linuxfoundation.org/legal/generative-ai).
- [x] I have the experience and knowledge necessary to answer maintainer questions about the content of this PR, without using AI.


eddie-knight (dismissed):

jpower432:
Noting a follow up to this work -- #498 captures expressing conditionals which is not addressed by this PR

cmd/internal/cmd/projection/cmd.go
```diff
@@ -11,7 +11,7 @@ import (
 var cue2OpenAPICmd = &cobra.Command{
 	Use:   "cue2openapi",
 	Short: "Convert CUE schema to OpenAPI YAML",
-	Long: `Convert CUE schema definitions to OpenAPI 3.0.3 YAML format.
+	Long: `Convert CUE schema definitions to OpenAPI 3.1.0 YAML format.
```
jpower432: Stayed at `3.0.3`
```suggestion
	Long: `Convert CUE schema definitions to OpenAPI 3.0.3 YAML format.
```

cmd/internal/cmd/diff/breakingcheck_integration_test.go
```diff
@@ -44,7 +44,7 @@ func TestBreakingCheckIntegration(t *testing.T) {
 
 	// Mutate the candidate by dropping a required entry from one schema. Removing
 	// a required property is consumer-breaking (the response can no longer promise
-	// it), so oasdiff must classify this as an ERR-level change.
+	// it), so diff must classify this as an ERR-level change.
```
jpower432: typo
```suggestion
	// it), so oasdiff must classify this as an ERR-level change.
```

cmd/internal/cmd/diff/breaking_test.go
```diff
@@ -53,7 +57,7 @@ func TestBreakingChanges(t *testing.T) {
 
 // TestBreakingChangesExemptsExperimental confirms that a schema carrying
 // x-status "experimental" is exempted from the gate: loadWrapped maps it to
-// oasdiff's x-stability-level "alpha", which the default (beta) threshold
+// diff's x-stability-level "alpha", which the default (beta) threshold
```
jpower432: typo
```suggestion
// oasdiff's x-stability-level "alpha", which the default (beta) threshold
```

cmd/internal/cmd/projection/annotate.go
```diff
@@ -0,0 +1,434 @@
...
+	if len(failed) > 0 {
+		sort.Strings(failed)
+		return nil, fmt.Errorf("cannot enumerate the fields of %d definition(s), so their "+
+			"deprecations would be silently dropped: %s", len(failed), strings.Join(failed, "; "))
+	}
+	return out, nil
+}
+
+// docText joins a value's doc comments exactly as the upstream encoder does in
+// encoding/projection's getDoc (build.go:222-236 in v0.15.4): raw comment text
```
jpower432: rename error
```suggestion
// encoding/openapi's getDoc (build.go:222-236 in v0.15.4): raw comment text
```

cmd/internal/cmd/projection/prep.go
```diff
@@ -0,0 +1,264 @@
...
+type prepInfo struct {
+	// CUE CONVERTER WORKAROUND: DroppedConditionals records the source position
+	// of every cross-field conditional removed by stripComprehensions. The
+	// current CUE OpenAPI projector cannot translate these, so the caller reports their omission
+	// so a newly added conditional announces that it will not reach the
+	// generated document, rather than vanishing silently.
+	DroppedConditionals []string
+}
+
+// prepare rewrites a CUE file into the subset cuelang.org/go/encoding/projection
```
jpower432: rename error
```suggestion
// prepare rewrites a CUE file into the subset cuelang.org/go/encoding/openapi
```

cmd/internal/cmd/projection/prep.go
```diff
@@ -0,0 +1,264 @@
...
+			// for: the walker returned "" for an *ast.Alias label and
+			// dropped the field.
+			if al, ok := n.Label.(*ast.Alias); ok {
+				if lbl, ok := al.Expr.(ast.Label); ok {
+					n.Label = lbl
+				}
+			}
+		case *ast.CallExpr:
+			// CUE CONVERTER WORKAROUND (cuelang.org/go, all versions through v0.18.0-alpha.1):
+			// encoding/projection panics on a time.Format() reachable inside a
```
jpower432: rename error
```suggestion
			// encoding/openapi panics on a time.Format() reachable inside a
```

jpower432 (commented):
Found a few typos caused by refactor/rename in IDE.

eddie-knight (approved):

PR #480, opened 2026-09-12
feat(ai-agent): route capability to detection through vectors and threats

eeee2345:
Follow-on to #459 (merged) and the discussion on #428. #459's body deferred "proposing the missing capabilities" to a separate change; this is that change. Rebased onto main after #459 landed.

Why three artifacts rather than one

The flat mapping in #459 can say that a category and a capability are related; it cannot say which threat connects them. ATR's detection categories are vectors, not threats: threatcatalog.cue defines a threat as an opportunity for negative impact, vectorcatalog.cue defines a vector as a method through which a threat is realized, and a detection category is the second of those. So the ten categories become a VectorCatalog, a ThreatCatalog references them, and a ControlCatalog is cut by when inspection happens rather than by which category runs.

What is in the change (counts are from the files)

- atr-attack-vector-catalog.yaml: 10 vectors in 3 groups.
- atr-threat-catalog.yaml: 6 threats in 3 groups, written at impact altitude; 14 threat-to-vector edges covering all 10 vectors. The row that justifies the layer is model-security, which reaches three different threats (behaviour extraction is disclosure, deserialization is execution escape, poisoned adaptation data is supply chain). A flat mapping cannot express that.
- atr-detection-control-catalog.yaml: 5 controls in 3 groups. Three are Active with three assessment requirements each; two are Draft with one requirement each, written as the MUST it will be and the reason it is not yet assessable carried in the recommendation, because a requirement that cannot fail is not a requirement. 11 requirements over 4 distinct applicability sets.
- ai-agent-capability-catalog.yaml (0.1.0 -> 0.2.0): adds a model-supply-chain group with two capabilities, CAP-MODEL-ARTIFACT-LOADING and CAP-TRAINING-DATA-INGESTION. Neither model artifact loading nor fine-tuning data ingestion happens in agent runtime, so they did not belong under the existing extensibility group.
- test/schema_test.go: the three new catalogs are added as schema validation cases.

Guideline references were measured rather than picked: each control carries the modal OWASP ASI and MITRE ATLAS references from the ATR rule directories it is implemented with, as counted on the ATR corpus at authoring time (2026-08-24). That is how context-exfiltration ended up on AML.T0057 rather than T0056, and skill-compromise and model-security on AML.T0010 and ASI04. The egress control cites no ASI entry because the modal one there (ASI01) names the injection that induces the disclosure rather than the disclosure itself, and citing it would overstate the fit.

Verification, on the rebased head 7d02b33 against current main

- make test: passes (schema validation suite, including the three new catalog cases).
- make lintinsights: clean.
- make breaking-check: passes against the v1.5.0 baseline (no schema files are touched; the change is example catalogs plus test cases).
- DCO sign-off present. No overlap with the files #459 changed.

One note on the commit message: it was written before #459 merged, so its last paragraph describes a then-current local cue incompatibility with main. That is no longer the case on this rebased head, and the checks above are the ones that apply now.


jpower432:
Hi @eeee2345. Thanks for this PR. Since the content is growing in size and artifact number, I am curious if we can start pointing to the canonical location for this material. Is this something you are still exploring?

eeee2345:
@jpower432 — apologies for the slow reply. Short answer: yes, and you already gave me the right model for it in #422 — Gemara defines the schema, the content lives with its owner. I agreed with that at the time and then kept adding artifacts here anyway, which is on me.

Concretely, where and how:

Canonical location — the ATR repo, alongside spec/category-registry/, which is the versioned list of the ten detection categories. The vector catalog in this PR is a direct projection of that registry, so co-locating them is what stops the two drifting. Right now they can drift, because the projection lives in your repo and its source of truth lives in mine. You are right about the size and the artifact count, and I think that drift is the sharper version of the same problem.

Publication — gemara-publish-action, OCI bundle to ghcr.io under the ATR org. ATR already has a working GHCR push in CI, so the auth path exists; the new piece is a root artifact for Assemble to walk. Keyless signing on source, promotion left off until there is a reason for a second registry.

Cadence — tied to the category registry, not to the rule corpus. The registry is a versioned spec artifact rather than something a routine commit touches: it has two commits in its whole history, created in May and touched once in July. Rule-level churn is daily and deliberately does not reach these artifacts, which is why the catalogs are cut at category level rather than per rule. The plan is to republish on ATR release tags; on the history so far that is a small number of moves a year, though four months is not much to extrapolate from.

What I would do with this PR — close it, land the catalogs in ATR, publish the bundle, then open a much smaller PR here that adds url to the mapping-references in the two artifacts already merged, so they resolve to the canonical bundle instead of depending on a file sitting in the same directory. Your #475 made that clean: url takes any scheme, so oci: is a first-class answer rather than a workaround.

The part that is your call and not mine: whether examples/ai-agent keeps one minimal worked example or becomes a pure pointer. My preference would be to keep the capability catalog and the mapping here as the worked example and host nothing further, but I will do it whichever way you want it.

#500 looks like the same principle from the other direction — the canonical artifact stays with its owner and consumers read a published release asset. If you would rather settle that contract first and have ATR follow whatever shape you land on, that ordering works for me. I am not asking to be added to any dispatch list; I would just match the published contract. Say the word and I will hold.

Housekeeping either way: the green on this PR is from 12 Sep and main has moved twice since, so treat it as stale rather than as a current pass. Happy to write the outcome up on #422 and close that out as well.

jpower432:
Thanks @eeee2345, this plan works for me. Adding inline replies below.

> 
> Canonical location — the ATR repo, alongside spec/category-registry/, which is the versioned list of the ten detection categories. The vector catalog in this PR is a direct projection of that registry, so co-locating them is what stops the two drifting. Right now they can drift, because the projection lives in your repo and its source of truth lives in mine. You are right about the size and the artifact count, and I think that drift is the sharper version of the same problem.

This make sense. Good point on the drift. It would be much simpler to mange the projection along side the registry.

> 
> Publication — gemara-publish-action, OCI bundle to ghcr.io under the ATR org. ATR already has a working GHCR push in CI, so the auth path exists; the new piece is a root artifact for Assemble to walk. Keyless signing on source, promotion left off until there is a reason for a second registry.

Just a note, that we are currently merging this action with the new [`grcli`](https://github.com/gemaraproj/grcli) tool. Intending to keep the action backward compatible though if your are still interested in published to GHCR primarily.

:point_up_2:  @eddie-knight @sonupreetam 

> What I would do with this PR — close it, land the catalogs in ATR, publish the bundle, then open a much smaller PR here that adds url to the mapping-references in the two artifacts already merged, so they resolve to the canonical bundle instead of depending on a file sitting in the same directory. Your #475 made that clean: url takes any scheme, so oci: is a first-class answer rather than a workaround.

Yep! This make sense.
> 
> The part that is your call and not mine: whether examples/ai-agent keeps one minimal worked example or becomes a pure pointer. My preference would be to keep the capability catalog and the mapping here as the worked example and host nothing further, but I will do it whichever way you want it.

Agree, we are using the existing examples as test fixtures so good to keep them here.

> 
> #500 looks like the same principle from the other direction — the canonical artifact stays with its owner and consumers read a published release asset. If you would rather settle that contract first and have ATR follow whatever shape you land on, that ordering works for me. I am not asking to be added to any dispatch list; I would just match the published contract. Say the word and I will hold.

No need to hold on #500. The action and CLI tool will verify your artifacts against the CUE schemas. We have the projection into JSON Schema and OpenAPI primarily to support codegen and documentation generation.

PR #481, opened 2026-09-12
test: add the missing negative fixture for the payload-or-source rule

astrogilda:
The payload or source rule added in #475 is the only recent audit-log constraint without a fixture that exercises it; every other one carries a bad-* file and a table row asserting the failure, so this rule's enforcement rests on reading the schema.

This adds the missing case: `good-audit-log.yaml` with source removed from one evidence entry, leaving neither alternative. It is a minimal mutation of the positive fixture, so the only difference between accept and reject is the field under test.

Checked both ways before opening: the fixture is rejected, the positive control still passes, and swapping a valid document into the negative slot makes the new row fail. The url scheme widening and the coordinate and entry-id combination also shipped in #475 without negative cases. Say so and I will add both in this PR.


jpower432:
Thanks for adding this @astrogilda! Just needed to make a small change to the PR title for the checks to pass.

jpower432:
@astrogilda Happy to approve this PR as is, but if you wanted to include the additional negative missing cases here, that works too.

astrogilda:
Thanks, Jennifer, I took the wider option. 587c187 adds three scheme negatives for url and uri plus a positive for coordinate with entry-id, each checked by disabling its rule.

jpower432 (approved):
LGTM. Thanks for the contribution!

Issue #170, opened 2025-10-30
Enhance multi-source evaluation support in Layer 4/5


jpower432:
## Objective

Currently there is a location in the `EvaluationPlan` to define information about the assessment tool in the metadata under `author`. If we update the `EvalautionPlan` to allow tools to be attached to certain assessment procedures, we can enable a workflow like that supports multi-source Evaluations.

- Layer 4 `EvaluationLog` preserves granular outcomes from individual tools/executors to maintain integrity.
- Layer 5 aggregates multiple `EvaluationLogs` from different tools, deconflicts evidence across tools using the granular results, and identifies findings for enforcement.
- Layer 4 `EvaluationPlan` guides this aggregation

## Diagram

```mermaid
flowchart TD
    subgraph L1["Layer 1"]
        G1["Guidance 1<br/>(e.g., NIST CSF)"]
        G2["Guidance 2<br/>(e.g., ISO 27001)"]
        G3["Guidance 3<br/>(e.g., PCI DSS)"]
        G4["...<br/>More Guidances"]
    end
    subgraph L2["Layer 2"]
        C1["Control Catalog 1<br/>(e.g., CIS Benchmarks)"]
        C2["Control Catalog 2<br/>(e.g., CCC)"]
        C3["Control Catalog 3<br/>(e.g., OSPS)"]
        C4["...<br/>More Controls"]
    end
    subgraph L3["Layer 3"]
        P1["Policy Document 1"]
        P2["Policy Document 2"]
        P3["...<br/>More Policies"]
    end

    subgraph L4["Layer 4"]
        EP["EvaluationPlan"]
        EL1["Evaluation Log 1"]
        EL2["Evaluation Log 2"]
        EL3["Evaluation Log 3"]
        EL4["...<br/>More Logs"]
    end

    subgraph L5["Layer 5"]
        F["Findings"]
        EA["EnforcementAction"]
        O["Layer5 Library"]
    end

    subgraph L6["Layer 6: Audit"]
        A["Audit"]
    end


    G1 --> P1
    G2 --> P1
    G3 --> P2
    G4 --> P3
    
    C1 --> P1
    C2 --> P1
    C3 --> P2
    C4 --> P3

    P1 --> EP
    P2 --> EP
    P3 --> EP

    EP --> EL1
    EP --> EL2
    EP --> EL3
    EP --> EL4

    EL1 -->|"technical status"|O
    EL2 -->|"technical status"|O
    EL3 -->|"technical status"|O
[...]

PR #482, opened 2026-09-12
feat: extend AssessmentPlan for multi-source evaluation

astrogilda:
Issue #170 asks for multi-source evaluation, and PR 236 already gave the plan half an owner. It removed the Layer 4 EvaluationPlan in favor of `assessment-plans` on the Layer 3 Policy, and AssessmentLog.plan points back at that. So this extends AssessmentPlan and adds no second artifact.

AcceptedMethod gains rank and environment-requirements. AssessmentPlan gains `conflict-resolution`, which a matchN constraint requires whenever a requirement names more than one evaluation method.

I have added five fixtures, four negative, each a minimal mutation of the positive; make test and make breaking-check against v1.5.0 pass on 107e629.

Four admission rails run this policy shape in CI, OPA rego, Kyverno CEL, Kyverno JMESPath and CUE: https://github.com/astrogilda/agent-evidence-admission

Edited 2026-09-16 after review: the first version added a separate evaluationplan.cue and an ArtifactType value.


jpower432 (changes requested):
Thanks for submitting this @astrogilda. I added some context on the originating issue (#170) to help move this to `Policy`. Would be great to get your thoughts.

astrogilda:
You're right, and PR 236 settles it: the Policy owns the plan, so d2a1673 moves this onto AssessmentPlan, keeping conflict-resolution explicit. An implementation default is one every implementer picks differently.

astrogilda:
@jpower432 I've pushed the AssessmentPlan move we discussed. The branch is rebased on main and CI is green, so I'd appreciate another look whenever you have a moment.

policy.cue
```diff
@@ -122,8 +139,51 @@ package gemara
...
+	// rank orders this method against the others on the same assessment plan; lower is
+	// higher precedence. It exists so that a rank-based conflict resolution has a total
+	// order to work with, and it is optional because a plan with one method needs none.
+	rank?: int & >=1
+
+	// environment-requirements states the environment this method is expected to run in.
+	// Its observed counterpart belongs on the EvaluationLog: a plan that states a required
+	// environment and a log that records none are not reconcilable, and the pair is the
+	// point of stating either.
+	"environment-requirements"?: #EnvironmentSpec @go(EnvironmentRequirements)
```
jpower432, line 152: @astrogilda The `environment-requirements` seems like it might be out of scope for this feature. Still something worth discussing more in my opinion, but doesn't seem to apply to log conflict resolution.

policy.cue
```diff
@@ -122,8 +139,51 @@ package gemara
 	required:     *false | bool
 	description?: string
 	executor?:    #Actor
+
+	// rank orders this method against the others on the same assessment plan; lower is
+	// higher precedence. It exists so that a rank-based conflict resolution has a total
+	// order to work with, and it is optional because a plan with one method needs none.
+	rank?: int & >=1
```
jpower432, line 146: I think we need  `rank` conditionally required. When a `rank` based conflict resolution strategy is selected to would be used to flag if there is not one set.

policy.cue
```diff
@@ -122,8 +139,51 @@ package gemara
...
+	"observation-source"?: string @go(ObservationSource)
+}
+
+// Digest is a cryptographic hash in algorithm:encoded form (e.g. sha256:abc123...).
+// The grammar is the one #EvidenceMapping.digest already uses inline, named here so the
+// two are the same shape by construction rather than by coincidence.
+#Digest: =~"^[a-z0-9]+(?:[+._-][a-z0-9]+)*:[a-zA-Z0-9=_-]+$"
+
+// ConflictPolicy states how disagreeing evaluation results are resolved.
+#ConflictPolicy: "highest-rank" | "unanimous" | "most-recent" | "escalate" @go(-)
```
jpower432, line 185: I think it would be helpful to have comments on these different options. `highest-rank` makes sense. Some questions those might answer?

What does `most-recent` look at? Most recent log (`metadata.date`)? Most recent `Assessment` run (`log.start`)?
Some the `methods` are required. Would `unanimous` only include required methods?
Does `escalate` mean the conflict could not be resolve? Would the resolved result be `Needs Review`?

jpower432, line 185: This is more of a nit, but `ConflictPolicy` collides a bit with `Policy`. Made a suggestion below.
```suggestion
#ConflictResolution: "highest-rank" | "unanimous" | "most-recent" | "escalate" @go(-)
```

jpower432 (changes requested):
Thanks for making the changes @astrogilda. I added a few suggestions and had one question about scope.

